PR #14, opened 2023-02-17
	modified:   QSymbolicsBase/src/QSymbolicsBase.jl

JackPostlewaite:
modified:   QSymbolicsBase/src/basic_ops_homogeneous.jl
	modified:   QSymbolicsBase/src/basic_ops_inhomogeneous.jl


Krastanov:
thanks for this! After minor cleanup I added it in PR #15

PR #20, opened 2023-07-23
add conditional clifford gates

Krastanov:


Krastanov:
@Barosandu , please check the commit messages and each commit that I added separately. I tried to comment on why things are being done and how it was done. It would be useful for you if you work on other improvements within the library.

Barosandu:
I will check it tomorrow. We returned from France today

PR #40, opened 2024-06-04
adding Commutators, Anticommutator, Daggers

apkille:
Defined constructors for bra, product of operators, commutator, anti commutator, and dagger. Also added scaling and a few simplification rules.


src/QSymbolicsBase/predefined.jl
```diff
@@ -205,23 +205,28 @@ function Base.show(io::IO, x::SProjector)
...
+"""Dagger, i.e., adjoint of quantum objects (kets, bras, operators, and inner products)"""
+@withmetadata struct SDagger{T<:Union{Complex, QObj}} <: Symbolic{Union{Complex, T}}
+    obj
 end
 istree(::SDagger) = true
 arguments(x::SDagger) = [x.ket]
 operation(x::SDagger) = dagger
 exprhead(x::SDagger) = :dagger
-dagger(x::Symbolic{AbstractKet}) = SDagger(x)
+dagger(x::Symbolic{AbstractKet}) = SBra(x.name, x.basis)
```
Krastanov, line 216: A ket might not be a simple named ket, rather some more sophisticated object. This method, as defined, makes sense for `SKet`, but not for any other symbolic ket types.

src/QSymbolicsBase/predefined.jl
```diff
@@ -205,23 +205,28 @@ function Base.show(io::IO, x::SProjector)
     print(io,"]")
 end
 
-"""Dagger a Ket into Bra."""
-@withmetadata struct SDagger <: Symbolic{AbstractBra}
-    ket::Symbolic{AbstractKet}
+"""Dagger, i.e., adjoint of quantum objects (kets, bras, operators, and inner products)"""
+@withmetadata struct SDagger{T<:Union{Complex, QObj}} <: Symbolic{Union{Complex, T}}
```
Krastanov, line 209: I do not think we should permit `Complex` as a type parameter here. Any reason for adding it?

src/QSymbolicsBase/rules.jl
```diff
@@ -1,3 +1,6 @@
+""" Predictate functions """
```
Krastanov, line 1: this docstring does not seem to describe the function it is attached to

Krastanov:
I left a couple of minor comments, nothing to interesting, and you might have already been thinking about some of these things.

I will convert this to a draft. When it is ready for review and it has all tests pass, just convert it back and request a review.

src/QSymbolicsBase/predefined.jl
apkille: My mistake, just fixed it.

apkille:
I updated functionalities of (anti)commutator, dagger, added operator properties (hermitian, unitary, and commutative), wrote corresponding simplification rules, which are all tested. I added doctests as well for basic operations.

src/QSymbolicsBase/objects.jl
```diff
@@ -0,0 +1,71 @@
...
+Base.show(io::IO, x::SKet) = print(io, "|$(symbollabel(x))⟩")
+Base.show(io::IO, x::SBra) = print(io, "⟨$(symbollabel(x))|")
+Base.show(io::IO, x::SOperator) = print(io, "$(symbollabel(x))")
+Base.show(io::IO, x::SymQObj) = print(io, symbollabel(x)) # fallback that probably is not great
+
+ishermitian(x::SOperator) = false
+isunitary(x::SOperator) = false
+iscommutative(x::SOperator) = false
+
+"""Inverse Operator"""
```
Krastanov, line 29: what is an "inverse operator"

src/QSymbolicsBase/objects.jl
```diff
@@ -0,0 +1,71 @@
...
+@withmetadata struct SInverseOperator <: Symbolic{AbstractOperator}
+    op::Symbolic{AbstractOperator}
+end
+isexpr(::SInverseOperator) = true
+iscall(::SInverseOperator) = true
+arguments(x::SInverseOperator) = [x.op]
+operation(x::SInverseOperator) = inverse
```
Krastanov, line 36: What is `inverse`? Is such a function declared anywhere?

If `inverse` does not exist we should probably figure out what tests to add so that issues like this one get caught automatically in the future.

src/QSymbolicsBase/objects.jl
Krastanov: Probably needs a docstring that can answer that questions on its own as well.

src/QSymbolicsBase/objects.jl
```diff
@@ -0,0 +1,71 @@
...
+head(x::SInverseOperator) = :inverse 
+children(x::SInverseOperator) = [:inverse, x.op]
+basis(x::SInverseOperator) = basis(x.op)
+Base.show(io::IO, x::SInverseOperator) = print(io, "$(x.op)⁻¹")
+inverse(x::Symbolic{AbstractOperator}) = SInverseOperator(x)
+
+"""Hermitian Operator"""
```
Krastanov, line 43: same question about what this is and about expanding docstring

src/QSymbolicsBase/objects.jl
```diff
@@ -0,0 +1,71 @@
...
+struct SHermitianOperator <: Symbolic{AbstractOperator}
+    op::Symbolic{AbstractOperator}
+end
+isexpr(::SHermitianOperator) = false
+basis(x::SHermitianOperator) = basis(x.op)
+Base.show(io::IO, x::SHermitianOperator) = print(io, "$(x.op)")
+hermitian(x::Symbolic{AbstractOperator}) = SHermitianOperator(x)
+ishermitian(::SHermitianOperator) = true
+
+"""Unitary Operator"""
```
Krastanov, line 53: same question about what this is and about expanding docstring

src/QSymbolicsBase/objects.jl
```diff
@@ -0,0 +1,71 @@
...
+inverse(x::Symbolic{AbstractOperator}) = SInverseOperator(x)
+
+"""Hermitian Operator"""
+struct SHermitianOperator <: Symbolic{AbstractOperator}
+    op::Symbolic{AbstractOperator}
+end
+isexpr(::SHermitianOperator) = false
+basis(x::SHermitianOperator) = basis(x.op)
+Base.show(io::IO, x::SHermitianOperator) = print(io, "$(x.op)")
+hermitian(x::Symbolic{AbstractOperator}) = SHermitianOperator(x)
```
Krastanov, line 50: Where does the function `hermitian` come from? What is it supposed to mean?

src/QSymbolicsBase/objects.jl
```diff
@@ -0,0 +1,71 @@
...
+ishermitian(::SHermitianOperator) = true
+
+"""Unitary Operator"""
+struct SUnitaryOperator <: Symbolic{AbstractOperator}
+    op::Symbolic{AbstractOperator}
+end
+isexpr(::SUnitaryOperator) = false
+basis(x::SUnitaryOperator) = basis(x.op)
+Base.show(io::IO, x::SUnitaryOperator) = print(io, "$(x.op)")
+unitary(x::Symbolic{AbstractOperator}) = SUnitaryOperator(x)
```
Krastanov, line 60: Where does the function `unitary` come from? What is it supposed to mean?

src/QSymbolicsBase/objects.jl
```diff
@@ -0,0 +1,71 @@
...
+isunitary(::SUnitaryOperator) = true
+
+"""Commutative Operator"""
+struct SCommutativeOperator <: Symbolic{AbstractOperator}
```
Krastanov, line 64: same question

src/QSymbolicsBase/objects.jl
```diff
@@ -0,0 +1,71 @@
...
+    op::Symbolic{AbstractOperator}
+end
+isexpr(::SCommutativeOperator) = false
+basis(x::SCommutativeOperator) = basis(x.op)
+Base.show(io::IO, x::SCommutativeOperator) = print(io, "$(x.op)")
+commutative(x::Symbolic{AbstractOperator}) = SCommutativeOperator(x)
```
Krastanov, line 70: same question

src/QSymbolicsBase/predefined.jl
```diff
@@ -102,22 +105,53 @@ isexpr(::OperatorEmbedding) = true
 @withmetadata struct XGate <: AbstractSingleQubitGate end
 eigvecs(g::XGate) = [X1,X2]
 symbollabel(::XGate) = "X"
+ishermitian(::XGate) = true
+isunitary(::XGate) = true
+iscommutative(::XGate) = false
```
Krastanov, line 110: "commutative" is not a property of a single linear operator. Only `const*I` commutes with all other operators for a general group. I do not think this can be consistently defined.

src/QSymbolicsBase/predefined.jl
```diff
@@ -207,25 +241,61 @@ function Base.show(io::IO, x::SProjector)
...
 isexpr(::SDagger) = true
 iscall(::SDagger) = true
-arguments(x::SDagger) = [x.ket]
+arguments(x::SDagger) = [x.obj]
 operation(x::SDagger) = dagger
 head(x::SDagger) = :dagger
-children(x::SDagger) = [:dagger, x.ket]
-dagger(x::Symbolic{AbstractKet}) = SDagger(x)
-dagger(x::SScaledKet) = SScaledBra(x.coeff, dagger(x.obj))
+children(x::SDagger) = [:dagger, x.obj]
```
Krastanov, line 272: With the switch to the newer version of the foundational dependencies, do we need both `arguments` and `children` or only one of them?

src/QSymbolicsBase/rules.jl
```diff
@@ -40,7 +47,83 @@ end
 
 FLATTEN_RULES = [
     @rule(~x::isnotflat_precheck(⊗) => flatten_term(⊗, ~x)),
-    @rule ⊗(~~xs::hasscalings) => prefactorscalings_rule(xs)
+    @rule ⊗(~~xs::hasscalings) => prefactorscalings_rule(xs) # Used to perform (a*|k⟩) ⊗ (b*|l⟩) → (a*b) * (|k⟩⊗|l⟩) 
 ]
 
 tensor_simplify = Fixpoint(Chain(FLATTEN_RULES))
+
+"""Quantum circuit identities"""
```
Krastanov, line 55: these probably should be called "Pauli" identities.

src/QSymbolicsBase/rules.jl
```diff
@@ -40,7 +47,83 @@ end
...
+_isYX(x) = x isa SApplyOp && x.terms == [Y,X]
+_isZY(x) = x isa SApplyOp && x.terms == [Z,Y]
+_isXZ(x) = x isa SApplyOp && x.terms == [X,Z]
+_isHXH(x) = x isa SApplyOp && x.terms == [H,X,H]
+_isHYH(x) = x isa SApplyOp && x.terms == [H,Y,H]
+_isHZH(x) = x isa SApplyOp && x.terms == [H,Z,H]
+
+
+CIRCUIT_RULES = [
+    @rule(~x::_isXX => I),
```
Krastanov, line 71: I would be very surprised if there isn't a simpler cleaner way to implement these. Creating a predicate to be used in a single rule seems much too cumbersome. This is exactly what pattern matching is supposed to simplify.

src/QSymbolicsBase/rules.jl
```diff
@@ -40,7 +47,83 @@ end
...
+"""Commutator identities"""
+_isXYcommutator(x) = x isa SCommutator && _isX(x.op1) && _isY(x.op2)
+_isYZcommutator(x) = x isa SCommutator && _isY(x.op1) && _isZ(x.op2)
+_isZXcommutator(x) = x isa SCommutator && _isZ(x.op1) && _isX(x.op2)
+_isYXcommutator(x) = x isa SCommutator && _isY(x.op1) && _isX(x.op2)
+_isZYcommutator(x) = x isa SCommutator && _isZ(x.op1) && _isY(x.op2)
+_isXZcommutator(x) = x isa SCommutator && _isX(x.op1) && _isZ(x.op2)
+
+COMMUTATOR_RULES = [
+    @rule(~x::_isXYcommutator => 2*im*Z),
```
Krastanov, line 96: Same comment: this should use pattern matching, not such complicated rules.

src/QSymbolicsBase/rules.jl
```diff
@@ -40,7 +47,83 @@ end
...
+_isZZanticommutator(x) = x isa SAnticommutator && _isZ(x.op1) && _isZ(x.op2)
+_isXYanticommutator(x) = x isa SAnticommutator && _isX(x.op1) && _isY(x.op2)
+_isYZanticommutator(x) = x isa SAnticommutator && _isY(x.op1) && _isZ(x.op2)
+_isZXanticommutator(x) = x isa SAnticommutator && _isZ(x.op1) && _isX(x.op2)
+_isYXanticommutator(x) = x isa SAnticommutator && _isY(x.op1) && _isX(x.op2)
+_isZYanticommutator(x) = x isa SAnticommutator && _isZ(x.op1) && _isY(x.op2)
+_isXZanticommutator(x) = x isa SAnticommutator && _isX(x.op1) && _isZ(x.op2)
+
+ANTICOMMUTATOR_RULES = [
+    @rule(~x::_isXXanticommutator => 2*I),
```
Krastanov, line 118: same comment

src/QSymbolicsBase/QSymbolicsBase.jl
```diff
@@ -17,24 +17,31 @@ import QuantumInterface:
...
+       SProjector,MixedState,IdentityOp,SInverseOperator,SHermitianOperator,SUnitaryOperator,SCommutativeOperator,
+       SApplyKet,SApplyBra,SApplyOp,SApplyOpSuper,SCommutator,SAnticommutator,SDagger,SBraKet,SOuterKetBra,
        HGate, XGate, YGate, ZGate, CPHASEGate, CNOTGate,
        XBasisState, YBasisState, ZBasisState,
        NumberOp, CreateOp, DestroyOp,
-       XCXGate, XCYGate, XCZGate, YCXGate, YCYGate, YCZGate, ZCXGate, ZCYGate, ZCZGate
+       XCXGate, XCYGate, XCZGate, YCXGate, YCYGate, YCZGate, ZCXGate, ZCYGate, ZCZGate, 
+       circuit_simplify, commutator_simplify, anticommutator_simplify, 
+       ishermitian, isunitary, iscommutative,
+       hermitian, unitary, commutative, inverse
```
Krastanov, line 44: you probably want `LinearAlgebra.inv`, not a brand new `inverse`

src/QSymbolicsBase/QSymbolicsBase.jl
```diff
@@ -151,12 +158,13 @@ end
...
 # TODO check that this does not cause incredibly bad runtime performance
 # use a macro to provide specializations if that is indeed the case
-propsequal(x,y) = all(n->getproperty(x,n)==getproperty(y,n), propertynames(x))
+propsequal(x,y) = all(n->isequal(getproperty(x,n),getproperty(y,n)), propertynames(x))
 
 ##
 # Most symbolic objects defined here
 ##
 
+include("objects.jl")
```
Krastanov, line 167: this probably needs more descriptive name. Something like `literals` or `base_litrals` or `literal_objects`

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -82,7 +87,73 @@ Base.show(io::IO, x::SAddOperator) = print(io, "("*join(map(string, arguments(x)
 const SAddBra = SAdd{AbstractBra}
 Base.show(io::IO, x::SAddBra) = print(io, "("*join(map(string, arguments(x)),"+")::String*")") # type assert to help inference
 
-"""Tensor product of quantum objects (kets, operators, or bras)."""
+# defines commutativity for addition of quantum objects
+const CommQObj = Union{SAddBra, SAddKet, SAddOperator}
```
Krastanov, line 91: What is this for? It does not seem to be used anywhere else.

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -82,7 +87,73 @@ Base.show(io::IO, x::SAddOperator) = print(io, "("*join(map(string, arguments(x)
...
+"""Symbolic application of operator on operator
+
+```jldoctest
+julia> A = SOperator(:A, SpinBasis(1//2)); B = SOperator(:B, SpinBasis(1//2));
+
+julia> A*B 
+AB
+```
+"""
+@withmetadata struct SApplyOp <: Symbolic{AbstractOperator}
```
Krastanov, line 126: I know this is following the naming convention from `ApplyKet` but here it does not seem to be as easy to follow. Maybe we can remove it to something like `SOpProduct` or something else easier to understand at a glance.

Krastanov: or `SMulOperator` to follow the convention from `SAddOperator`

src/QSymbolicsBase/objects.jl
Krastanov: on reading the rest of the PR, I think I am getting a better idea of what this is (not so much for the other 3). This does not seem like a basic literal. Probably it should go wherever the dagger is defined.

src/QSymbolicsBase/rules.jl
Krastanov: well, I checked, and I am indeed very surprised how convoluted it is, but it can still be simpler (and more "pattern match"-ie)  than this version.

E.g. consider this (misusing `tensor` instead of multiplication because the current version of QSymbolics does not have an operation multiplication implemented)

```
julia> _isa(T) = x->isa(x,T)
_isa (generic function with 1 method)

julia> r = @rule ~O₁::_isa(XGate) ⊗ ~O₂::_isa(ZGate) => -im*Y
~(O₁::_isa(XGate)) ⊗ ~(O₂::_isa(ZGate)) => -im * Y

julia> r(X ⊗ Z)
0 - 1imY
```

src/QSymbolicsBase/rules.jl
Krastanov: given my other comment, this should be as simple as `@rule commutator(~o1::_isa(XGate), ~o2::_isa(Ygate)) => 2im Z`

Krastanov (changes requested):
This is a very solid start and I am very happy to see the momentum you already have!

I have a ton of questions for now, that will inform me how to understand this work. Let's go through them and then we can do another round of review and changes.

Converting back to draft -- convert to nondraft and request a review when you feel it is appropriate.

src/QSymbolicsBase/objects.jl
apkille: Sorry, it's a terse docstring, I'll fix that. I defined `SInverseOperator` so one could have the inverse of a symbolic operator with the `inverse` function.

src/QSymbolicsBase/objects.jl
apkille: The idea with Hermitian and unitary operators was that one could make a symbolic operator hermitian or unitary with the `hermitian` and `unitary` functions, respectively. Maybe there's a more straightforward way to do this, or it would be clear if I just took the time to expand the docstrings.

src/QSymbolicsBase/predefined.jl
apkille: My bad, silly blunder of mine. I'll remove this property.

src/QSymbolicsBase/predefined.jl
apkille: According to most recent `TermInterface` docs, when `isexpr` is true, both `head` and `children` must be defined. Similarly, `operations` and `arguments` must be defined when `iscall` is true. If `iscall` is true, then `isexpr` must be true.

src/QSymbolicsBase/objects.jl
Krastanov: oh, I see now. Specifically, you have made them `isexpr(::SHermitianOperator) = false`. That makes a lot of sense.

I think I would still prefer if they do not wrap around `SOperator`. For starters, even if we keep the structure you have, `op` should probably be only literal, not arbitrary, i.e. `op::SOperator`.

But I would prefer if there is no wrapping at all, i.e. `SHermitianOperator` and `SOperator` should be basically the same, except for `ishermitian` being defined for the former.

`SInvOperator` should probably be together with `SDagger` though, not here, as it is not a literal.

apkille: OK, I agree with all of this, will make the updates.

Was also wondering if we wanted to define an operator that was both hermitian and unitary. Should we simply define another structure, say `SHermitianUnitaryOperator`?

src/QSymbolicsBase/rules.jl
apkille: That code doesn't work for me: what you wrote is very similar to the example I showed you on Tuesday. When I define that rule, `r(X⊗Z)` gives `nothing`.

src/QSymbolicsBase/objects.jl
Krastanov: the "proper" way to do this would be to either:

- Use "traits" or a type parameter or something in the metadata, but that would require a bit more design work
- Set up an "assumptions" system... which is one of the most difficult things for any computer algebra system. The design discussions on topics like this span megabytes of email threads on any dev mailing list (e.g. sympy's)

I would prefer if you spend your time on the lower-hanging fruits, so I would suggest indeed going for `SHermitianUnitaryOperator`. That can always be refactored in a non-breaking manner to use one of the aforementioned "better" approaches.

src/QSymbolicsBase/rules.jl
Krastanov: weird... on this exact branch I get the following (copying the status of installed projects for completeness)

```
(quantumjulia) pkg> st
Status `~/Documents/ScratchSpace/quantumjulia/Project.toml`
⌅ [c3fe647b] AbstractAlgebra v0.40.9
  [cbdf2221] AlgebraOfGraphics v0.6.19
  [2c1a90cd] BPGates v1.1.0 `BPGates.jl`
  [6e4b80f9] BenchmarkTools v1.5.0
  [824d6782] Bonito v3.1.1
  [13f3f980] CairoMakie v0.12.2
  [9961bab8] Cbc v1.2.0
  [861a8166] Combinatorics v1.0.2
  [6ed1e86c] ConcurrentSim v1.4.1 `ConcurrentSim.jl`
  [992eb4ea] CondaPkg v0.2.22
  [a93c6f00] DataFrames v1.6.1
  [864edb3b] DataStructures v0.18.20
  [459566f4] DiffEqCallbacks v3.6.2
  [31c24e10] Distributions v0.25.109
  [ffbed154] DocStringExtensions v0.9.3
  [e30172f5] Documenter v1.4.1
  [5789e2e9] FileIO v1.16.3
  [e9467ef8] GLMakie v0.10.2
  [c27321d9] Glob v1.3.1
  [1ecd5474] GraphMakie v0.5.11
  [86223c79] Graphs v1.11.1
  [06909019] GraphsFlows v0.1.1
  [c3af3a8c] GraphsMatching v0.2.0
  [cd3eb016] HTTP v1.10.8
  [7073ff75] IJulia v1.24.2
  [c3a54625] JET v0.9.3
  [0f8b85d8] JSON3 v1.14.0
  [4076af6c] JuMP v1.22.1
  [98e50ef6] JuliaFormatter v1.0.56
  [3c486d74] LDPCDecoders v0.3.1 `LDPCDecoders.jl`
  [b964fa9f] LaTeXStrings v1.3.1
  [1914dd2f] MacroTools v0.5.13
  [ee78f7c6] Makie v0.21.2
⌅ [2edaba10] Nemo v0.43.3
  [46757867] NetworkLayout v0.4.6
  [8913a72c] NonlinearSolve v3.12.4
  [f1435218] Oscar v1.0.3
  [df9a0d86] Oxygen v1.5.10
  [e4faabce] PProf v3.1.0
  [2ae35dd2] Permutations v0.4.21
  [32113eaa] PkgBenchmark v0.2.12
  [91a5bcdd] Plots v1.40.4
  [21216c6a] Preferences v1.4.3
  [92933f4c] ProgressMeter v1.10.0
  [438e738f] PyCall v1.96.4
  [17f5de1a] PyQDecoders v0.2.0 `PyQDecoders.jl`
  [6099a3de] PythonCall v0.9.20
  [b0d11df0] Quantikz v1.3.1 `Quantikz.jl`
  [0525e862] QuantumClifford v0.9.4 `QuantumClifford.jl`
  [5717a53b] QuantumInterface v0.3.4 `QuantumInterface.jl`
  [6e0679c1] QuantumOptics v1.0.16 `QuantumOptics.jl`
  [4f57444f] QuantumOpticsBase v0.4.22 `QuantumOpticsBase.jl`
  [efa7fd63] QuantumSymbolics v0.3.0 `QuantumSymbolics.jl`
  [c5292f4c] ResumableFunctions v0.6.10 `ResumableFunctions.jl`
  [295af30f] Revise v3.5.14
  [276daf66] SpecialFunctions v2.4.0
  [860ef19b] StableRNGs v1.0.2
  [856f2bd8] StructTypes v1.10.0
  [8e1ec7a9] SumTypes v0.5.8 `~/.julia/dev/SumTypes`
  [1b6eb727] SwaggerMarkdown v0.2.1
  [24249f21] SymPy v2.1.1
  [0c5d862f] Symbolics v5.30.3
  [276b4fcb] WGLMakie v0.10.2
  [ddb6d928] YAML v0.4.11
  [37e2e46d] LinearAlgebra
  [56ddb016] Logging
  [8dfed614] Test
Info Packages marked with ⌅ have new versions available but compatibility constraints restrict them from upgrading. To see why use `status --outdated`

julia> using QuantumSymbolics, Symbolics

julia> _isa(T) = x->isa(x,T)
_isa (generic function with 1 method)

julia> r = @rule ~O₁::_isa(XGate) ⊗ ~O₂::_isa(ZGate) => -im*Y
~(O₁::_isa(XGate)) ⊗ ~(O₂::_isa(ZGate)) => -im * Y

julia> r(X ⊗ Z)
0 - 1imY

julia> versioninfo()
Julia Version 1.10.4
Commit 48d4fd48430 (2024-06-04 10:41 UTC)
Build Info:
  Official https://julialang.org/ release
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 32 × AMD Ryzen 9 7950X 16-Core Processor
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, znver3)
Threads: 32 default, 1 interactive, 16 GC (on 32 virtual cores)
```

src/QSymbolicsBase/basic_ops_homogeneous.jl
apkille: During my tests, commutativity for addition did not always hold, so I added an `isequal` method for adding operations. Take the following example:
```
julia> A = SOperator(:A, SpinBasis(1//2)); B = SOperator(:B, SpinBasis(1//2));

julia> dagger(A+B)
(A†+B†)

julia> dagger(A+B)
(B†+A†)
```
For reasons that I'm still trying to determine, `dagger(A+B)` flips between the ordering of the operators when it simplifies. This also happens with the commutators when you expand them. I could spend time working on ordering objects in expressions (as is done in `SymbolicUtils`) to make sure this doesn't happen. That being said, before I defined `isequal` for adding operations, the test would break because `(A†+B†)` was not equal to `(B†+A†)`, which obviously isn't what we want. So, to at least make sure addition of objects are commutative, I added an additional `isequal` method.

src/QSymbolicsBase/rules.jl
apkille: Ah, I wasn't using the latest version of `Symbolics`. Works now, thanks!

Krastanov: oh, if that was a source of things not working, we should appropriately update the compat in Project.tom.

src/QSymbolicsBase/predefined.jl
apkille: So to answer your question, for most cases, yes.

src/QSymbolicsBase/basic_ops_inhomogeneous.jl
```diff
@@ -42,31 +73,38 @@ operation(x::SBraKet) = *
...
 children(x::SBraKet) = [:*,x.bra,x.ket]
 Base.:(*)(b::Symbolic{AbstractBra}, k::Symbolic{AbstractKet}) = SBraKet(b,k)
-function Base.show(io::IO, x::SBraKet)
-    print(io,x.bra)
-    print(io,x.ket)
-end
+Base.show(io::IO, x::SBraKet) = begin print(io,x.bra); print(io,x.ket) end
 
 """Symbolic application of a superoperator on an operator"""
-@withmetadata struct SApplyOp <: Symbolic{AbstractOperator}
```
Krastanov, line 51: This being renamed worries me. Any specific reason not to stick to the previous name? It is not a public name, so it is not too breaking, but I want to have a good reason to do it.

src/QSymbolicsBase/predefined.jl
```diff
@@ -90,6 +90,9 @@ basis(::AbstractSingleQubitGate) = qubit_basis
 basis(::AbstractTwoQubitGate) = qubit_basis⊗qubit_basis
 Base.show(io::IO, x::AbstractSingleQubitOp) = print(io, "$(symbollabel(x))")
 Base.show(io::IO, x::AbstractTwoQubitOp) = print(io, "$(symbollabel(x))")
+Base.:(*)(xs::AbstractSingleQubitGate...) = pauli_simplify(SMulOperator(collect(xs)))
```
Krastanov, line 93: I would prefer if we avoid complex (potentially expensive, traversing the entire tree deep into the leafs of the arguments) simplification routines in constructors of symbolic expressions. Only extremely basic simplifications should be done automatically without the user requesting them.

An issue like this led to the sympy team having to create a whole new set of constructors without simplification because it made the rest of the codebase too difficult to support otherwise.

src/QSymbolicsBase/predefined.jl
```diff
@@ -90,6 +90,9 @@ basis(::AbstractSingleQubitGate) = qubit_basis
...
+commutator(o1::AbstractSingleQubitGate, o2::AbstractSingleQubitGate) = commutator_simplify(SCommutator(o1, o2))
+anticommutator(o1::AbstractSingleQubitGate, o2::AbstractSingleQubitGate) = anticommutator_simplify(SAnticommutator(o1, o2))
```
Krastanov, line 95: same comment about automatic simplification

test/test_express_opt.jl
```diff
@@ -35,4 +35,4 @@ op = N⊗X
 @test express(op*state) ≈ express(op)*express(state)
 @test express(op*state) ≈ express(F1⊗Z2)
 
-state = (3im*(2*dagger(Z1)+dagger(Y1))) * (3im*(2*X1+X2))
```
Krastanov, line 38: Why is this removed from the tests?

Krastanov (changes requested):
Looks great! A few minor things left to polish and it should be ready for merging.

src/QSymbolicsBase/QSymbolicsBase.jl
```diff
@@ -127,12 +133,12 @@ newwithmetadata(x) = x
...
 ##
 
-const QObj = Union{AbstractKet,AbstractOperator,AbstractSuperOperator,AbstractBra}
+const QObj = Union{AbstractBra,AbstractKet,AbstractOperator,AbstractSuperOperator}
 const SymQObj = Symbolic{<:QObj} # TODO Should we use Sym or Symbolic... Sym has a lot of predefined goodies, including metadata support
 Base.:(-)(x::SymQObj) = (-1)*x
 Base.:(-)(x::SymQObj,y::SymQObj) = x + (-y)
 
-function Base.isequal(x::X,y::Y) where {X<:SymQObj, Y<:SymQObj}
+function Base.isequal(x::X,y::Y) where {X<:Union{SymQObj, Symbolic{Complex}}, Y<:Union{SymQObj, Symbolic{Complex}}}
```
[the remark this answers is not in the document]
apkille: Inner products are complex: `SBraKet <: Symbolic{Complex}`.

src/QSymbolicsBase/basic_ops_inhomogeneous.jl
apkille: My worry is that `SApplyOp` sounds like it is symbolic multiplication of operators, rather than application of a superoperator on an operator.

src/QSymbolicsBase/predefined.jl
apkille: OK, how about creating a simplify function for the user to manually simplify an expression?

Krastanov: Do keep the smaller simplify functions you already have (up to potentially a rename) -- it is nice to have easy-to-access single-purpose simplification routines.

src/QSymbolicsBase/QSymbolicsBase.jl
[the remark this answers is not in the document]
Krastanov: noted, marking as resolved

src/QSymbolicsBase/basic_ops_inhomogeneous.jl
Krastanov: You are right about that. How about `SSuperOpApply`? My goal is to distinguish it from the others who follow the convention `SApply<the thing being acted upon>` and not `SApply<the thing that is acting>`.

src/QSymbolicsBase/predefined.jl
apkille: Yes, there is a `simplify` exported by Symbolics. `simplify_quantum` sounds OK, but is long. Maybe `qsimplify`?

Krastanov: sounds good!

test/test_express_opt.jl
apkille: There is currently no predefined dagger for any of the basis states.

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -82,7 +87,73 @@ Base.show(io::IO, x::SAddOperator) = print(io, "("*join(map(string, arguments(x)
...
+        end
+    end
+    false
+end
+function Base.isequal(x::CommQObj, y::CommQObj)
+    if typeof(x)==typeof(y)
+        if isexpr(x)
+            if operation(x)==operation(y)
+                ax,ay = arguments(x),arguments(y)
+                (length(ax) == length(ay)) && all(x -> _in(x, y), ax)
```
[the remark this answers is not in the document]
apkille: `Set` doesn't appear to work here, as equality of sets relies on `==` rather than `isequal` function. Here's an example:
```
julia> A = SOperator(:A, SpinBasis(1//2)); B = SOperator(:B, SpinBasis(1//2));

julia> Set([A+B]) == Set([B+A])
false

julia> isequal(A+B, B+A)
true

julia> A+B == B+A
false
```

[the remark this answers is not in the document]
apkille: I defined set containment of quantum objects with `_in(x::SymQObj, y::SymQObj)`. Performance-wise, it's probably not better, but unless we want to redefine `==` for quantum objects, it might be better to stick with this option.

test/test_express_opt.jl
Krastanov: But this worked previously, right? Isn't there still just a generic dagger wrapper? One should be able to represent `dagger(ket)` without having to be able to construct a bra. There is no need for predefined daggers of basis states.

Krastanov: generally things that work should not be broken -- that would require a much more careful release to make sure that other dependent libraries do not break on such a change

apkille: I agree, just made the change. Everything should be in working order.

src/QSymbolicsBase/rules.jl
```diff
@@ -1,18 +1,36 @@
...
-function prefactorscalings(xs)
+function prefactorscalings(xs; scalar=false) # If the scalar keyword is true, then only scalar factors will be returned as coefficients
     terms = []
     coeff = 1::Any
     for x in xs
         if isexpr(x) && operation(x) == *
             c,t = arguments(x)
-            coeff *= c
-            push!(terms,t)
+            if scalar == false
```
Krastanov, line 25: minor stylistic note, it is a bit more natural to write `if !scalar`. And `if scalar` below.

src/QSymbolicsBase/rules.jl
```diff
@@ -38,9 +56,84 @@ function isnotflat_precheck(*)
...
+For performance or single-purpose motivations, the user has the option to define a specific rewriter for `qsimplify` to apply to the expression.
+
+```jldoctest
+julia> qsimplify(anticommutator(σˣ, σˣ), rewriter=qsimplify_anticommutator)
+2𝕀
+```
+"""
+function qsimplify(s; rewriter=nothing)
+    if QuantumSymbolics.isexpr(s)
+        if rewriter == nothing
```
Krastanov, line 130: another stylistic note, it is more idiomatic to write `isnothing(rewriter)` or `rewriter === nothing`. Usually this does not matter because the compiler does "constant propagation" and basically ends up with this anyway, but it is a good habit to keep because `isnothing` and `===` guarantee the checks get compiled away when possible. The term to google about this is `type stability` and `specialization`

Krastanov (changes requested):
I think there are 3 very minor correctness issues and the rest looks great!

src/QSymbolicsBase/predefined.jl
```diff
@@ -207,25 +229,94 @@ function Base.show(io::IO, x::SProjector)
...
 operation(x::SDagger) = dagger
 head(x::SDagger) = :dagger
-children(x::SDagger) = [:dagger, x.ket]
-dagger(x::Symbolic{AbstractKet}) = SDagger(x)
-dagger(x::SScaledKet) = SScaledBra(x.coeff, dagger(x.obj))
+children(x::SDagger) = [:dagger, x.obj]
+dagger(x::Symbolic{AbstractBra}) = SDagger{AbstractKet}(x)
+dagger(x::Symbolic{AbstractKet}) = SDagger{AbstractBra}(x)
+dagger(x::Symbolic{AbstractOperator}) = SDagger{AbstractOperator}(x)
+dagger(x::SKet) = SBra(x.name, x.basis)
```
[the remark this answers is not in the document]
apkille: OK, suppose we have a symbolic ket, say `|a⟩`, and we want to apply `dagger` to this ket. So instead of following the definition that `dagger(|a⟩) == ⟨a|`, as done in this PR, we should leave it as `dagger(|a⟩) == |a⟩†`? And this is because we are kind of cheating with the former approach, rather than traversing the expression tree? I'm not disagreeing with you, I'm just making sure I'm understanding what you're saying.

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -76,13 +83,61 @@ Base.:(+)(xs::Vararg{Symbolic{<:QObj},0}) = 0 # to avoid undefined type paramete
 basis(x::SAdd) = basis(first(x.dict).first)
 
 const SAddKet = SAdd{AbstractKet}
-Base.show(io::IO, x::SAddKet) = print(io, "("*join(map(string, arguments(x)),"+")::String*")") # type assert to help inference
+function Base.show(io::IO, x::SAddKet)
+    ordered_terms = sort([repr(i) for i in arguments(x)])
```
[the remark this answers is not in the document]
Krastanov: and sticking to your `repr` seems to be appropriate. You can just remove `string`

src/QSymbolicsBase/predefined.jl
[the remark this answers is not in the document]
Krastanov: Yup, that is it. There is no guarantee that `|a⟩` and `⟨a|` are related at all (except for you implicitly deciding so). To make it explicit you need to write some pattern matching rule that makes it evident that they are related. Such a pattern matching rule would be difficult in a setup that is not based on the expressions API. By just creating a new expression type `dagger`, that problem of implicit relationships disappears -- everything becomes explicit.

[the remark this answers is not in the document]
apkille: Got it, just made changes.

Krastanov (approved):

Krastanov:
Woohoo!

apkille:
Excellent! 😁

Issue #46, opened 2024-06-18
add SZeroOperator, SZeroBra, SZeroKet structs


apkille:
```
julia> A = SOperator(:A, SpinBasis(1//2)); b = SBra(:b, SpinBasis(1//2)); k = SKet(:k, SpinBasis(1//2));

julia> @show 0*A 0*b 0*k;
0A = 0A
0 * b = 0⟨b|
0k = 0|k⟩
```

PR #50, opened 2024-06-20
Addresses #46 and fixes #47

### LINKED ISSUE #47: use set equality when checking for expression equality of commutative terms (and potentially generally use sets for commutative terms)
we already have

```
julia> Set([X1, X2]) == Set([X1, X2])
true

julia> Set([X1, X2]) == Set([X1, Z1])
false
```

Set uses hashing for quick (potentially false positive) equality check in inclusion (`log(n)`), followed by `isequal` check to avoid false positive. Thus set equalities have cost of `n log(n)` instead of `n^2` without the use of a smart datastructure.

Initially there was a worry that Set uses `==`, which would have broken its use with Symbolics (where `==` is used to create a symbolic equation), but that does not seem to be the case.

To avoid creating a new `Set` at every equality check, we should probably have it as part of the structure. E.g.

```
@withmetadata struct SAdd{T<:QObj} <: Symbolic{T}
    dict
    _set_precomputed
    _arguments_precomputed
end
```

It will take a bit more memory (not much given that the objects are not copied, only referenced), but it should speed up a lot of other code.

pinging @apkille as we had been discussing this in #40

apkille:
Added zero structs (#46) and created a hashing method for expressions containing quantum objects #47 in order to define expression equality for commutative operations.


src/QSymbolicsBase/QSymbolicsBase.jl
```diff
@@ -137,25 +138,14 @@ const QObj = Union{AbstractBra,AbstractKet,AbstractOperator,AbstractSuperOperato
 const SymQObj = Symbolic{<:QObj} # TODO Should we use Sym or Symbolic... Sym has a lot of predefined goodies, including metadata support
 Base.:(-)(x::SymQObj) = (-1)*x
 Base.:(-)(x::SymQObj,y::SymQObj) = x + (-y)
+Base.hash(x::SymQObj, h::UInt) = isexpr(x) ? hash(arguments(x), h) : hash([typeof(x),basis(x)], h)
```
Krastanov, line 141:
```suggestion
Base.hash(x::SymQObj, h::UInt) = isexpr(x) ? hash((head(x),arguments(x)), h) : hash((typeof(x),basis(x)), h)
```

The change to the first branch is to make the hash of `X1+X2` and `tensor(X1, X2)` different.

The change to the second branch is less interesting: in julia tuples are much easier for the compiler to optimize away (and not allocate) than vectors. The reason is that tuples are immutable, while vectors potentially can even change size, so tuples can go on the stack, but vectors have to go on the heap. A smarter compiler can in-principle fix that (There are some interesting experimental PR for this at the julia repo, in case you are interested. More generally, modifying the compiler from inside a user session is very easy in julia, so it is a good language if you want to study compiler optimization techniques).

By the way, could we add something more to the second branch so that `SKet(:x)` and `SKet(:y)` also have different tags? Maybe `symbollabel`?

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -29,9 +30,22 @@ arguments(x::SScaled) = [x.coeff,x.obj]
 operation(x::SScaled) = *
 head(x::SScaled) = :*
 children(x::SScaled) = [:*,x.coeff,x.obj]
-Base.:(*)(c, x::Symbolic{T}) where {T<:QObj} = SScaled{T}(c,x)
-Base.:(*)(x::Symbolic{T}, c) where {T<:QObj} = SScaled{T}(c,x)
-Base.:(/)(x::Symbolic{T}, c) where {T<:QObj} = SScaled{T}(1/c,x)
+function Base.:(*)(c, x::Symbolic{T}) where {T<:QObj} 
+    if iszero(c) || isa(x,SymZeroObj)
+        first(filter(i->i<:Symbolic{T}, Base.uniontypes(SymZeroObj)))()
```
Krastanov, line 35: If we make `SZero` a parametric type, then `SZero{T}()` can be a much simpler (and static) alternative to this filter expression.

src/QSymbolicsBase/QSymbolicsBase.jl
```diff
@@ -137,25 +138,14 @@ const QObj = Union{AbstractBra,AbstractKet,AbstractOperator,AbstractSuperOperato
...
 
-function _in(x::SymQObj, y::SymQObj)
-    for i in arguments(y)
-        if isequal(x, i)
-            return true
-        end
-    end
-    false
-end
 function Base.isequal(x::X,y::Y) where {X<:Union{SymQObj, Symbolic{Complex}}, Y<:Union{SymQObj, Symbolic{Complex}}}
```
Krastanov, line 143: I did not notice this in the previous PR, but I think this causes piracy because it will be the method to use when `isequal(::Symbolic{Complex},::Symbolic{Complex})` is called and we do not own neither `Symbolic` nor `Complex`.

To avoid that piracy (which get's detected by the `Aqua` test but it is currently turned off because there are a lot of other Piracies too), I suggest this gets renamed to `_isequal` without any type constraints and then two separate `Base.isequal` get created:

`Base.isequal(x::Union{...}, y::SymQObj) = _isequal(x, y)`
`Base.isequal(x::SymQObj, y::::Union{...}) = _isequal(x, y)`

That way you do not risk dispatching on this if you have two `Complex` arguments.

Krastanov: Which we also want to avoid because this will anyway fail if given two sums of complex scalars, because the complex sum probably does not have `_set_precomputed`

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -29,9 +30,22 @@ arguments(x::SScaled) = [x.coeff,x.obj]
...
+    else
+        SScaled{T}(c, x)
+    end
+end
+function Base.:(*)(x::Symbolic{T}, c) where {T<:QObj}
```
Krastanov, line 40: to avoid code repetition you can just write `Base.:(*)(x::Symbolic{T}, c) = c*x`

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -29,9 +30,22 @@ arguments(x::SScaled) = [x.coeff,x.obj]
...
+    end
+end
+function Base.:(*)(x::Symbolic{T}, c) where {T<:QObj} 
+    if iszero(c) || isa(x,SymZeroObj)
+        first(filter(i->i<:Symbolic{T}, Base.uniontypes(SymZeroObj)))()
+    else
+        SScaled{T}(c, x)
+    end
+end
+Base.:(/)(x::Symbolic{T}, c) where {T<:QObj} = iszero(c) ? error("cannot divide by zero") : SScaled{T}(1/c,x)
```
Krastanov, line 47:
```suggestion
Base.:(/)(x::Symbolic{T}, c) where {T<:QObj} = iszero(c) ? throw(DomainError(c,"cannot divide QSymbolics expressions by zero")) : SScaled{T}(1/c,x)
```

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -70,15 +84,20 @@ julia> k₁ + k₂
...
+arguments(x::SAdd) = x._arguments_precomputed
 operation(x::SAdd) = +
 head(x::SAdd) = :+
-children(x::SAdd) = [:+,SScaledKet(v,k) for (k,v) in pairs(x.dict)]
-Base.:(+)(xs::Vararg{Symbolic{T},N}) where {T<:QObj,N} = SAdd{T}(countmap_flatten(xs, SScaled{T}))
+children(x::SAdd) = [:+; x._arguments_precomputed]
+function Base.:(+)(xs::Vararg{Symbolic{T},N}) where {T<:QObj,N} 
+    nonzero_terms = filter!(x->!isa(x,SymZeroObj),collect(xs))
+    isempty(nonzero_terms) ? xs[1] : SAdd{T}(countmap_flatten(nonzero_terms, SScaled{T}), Set(collect(xs)), collect(xs))
+end
```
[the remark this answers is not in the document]
Krastanov: also, `xs[1]` might not exist anymore by the time you get to this line, so probably you should start with `xs = collect(xs); f = first(xs)`

src/QSymbolicsBase/predefined.jl
```diff
@@ -264,10 +265,9 @@ dagger(x::Symbolic{AbstractBra}) = SDagger{AbstractKet}(x)
 dagger(x::Symbolic{AbstractKet}) = SDagger{AbstractBra}(x)
 dagger(x::Symbolic{AbstractOperator}) = SDagger{AbstractOperator}(x)
 dagger(x::SScaledKet) = SScaledBra(conj(x.coeff), dagger(x.obj))
-dagger(x::SAddKet) = SAddBra(Dict(dagger(k)=>v for (k,v) in pairs(x.dict)))
+dagger(x::SAdd) = (+)([dagger(i) for i in arguments(x)]...)
```
Krastanov, line 268: not very sure about my suggestion here, but could you switch this to a generator comprehension instead of a list comprehension -- the compiler has a better chance of avoiding allocations with lazy generators.

Krastanov (dismissed):
Wonderful job! I have mostly stylistic/idioms comments. But I think I also found a place where we have introduced a small bug previously.

src/QSymbolicsBase/QSymbolicsBase.jl
apkille: I might be misunderstanding you, but won't your approach of creating two separate `Base.isequal` methods create ambiguity? For instance, if you are simply evaluating equality of a symbolic operator, then it's unclear which specific method to use. Just to clarify, when you say `Union{...}` you mean `Union{SymQObj, Symbolic{Complex}`, right?

Krastanov: You are right, that would create an ambiguity, this was a bad suggestion.

Better suggestion: just remove the Complex from this one and define `isequal(::SymQObj, ::Symbolic{Complex})=false` and the other order of arguments

apkille: I agree with this, but this still doesn't help when we're defining equality between inner products. How about we additionally define `Base.isequal(x::SBraKet, y::SBraKet) = isequal(x.bra, y.bra) && isequal(x.ket, y.ket)`?

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -70,15 +84,20 @@ julia> k₁ + k₂
 """
 @withmetadata struct SAdd{T<:QObj} <: Symbolic{T}
     dict
-    SAdd{S}(d) where S = length(d)==1 ? SScaled{S}(reverse(first(d))...) : new{S}(d)
+    _set_precomputed
+    _arguments_precomputed
+    SAdd{S}(d,s,a) where S = length(d)==1 ? SScaled{S}(reverse(first(d))...) : new{S}(d,s,a)
```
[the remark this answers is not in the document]
apkille: I agree the constructor was confusing. Now, we have an outer constructor that precomputes arguments and doesn't confusingly compute `reverse(first(d))...`.

src/QSymbolicsBase/QSymbolicsBase.jl
Krastanov: yup, that sounds necessary

There are other ones that might be necessary too, but let's not worry about them until they pop up

Krastanov:
A day or two ago I spent some time to make sure that all tests pass across this github organization, so I will be a bit more strict on tests during review. Here is a list of explanations:

The macos error was just a random happenstance

The codecov coverage issue is not great, but not a blocker

The performance tracking is not set up yet, that is why it is failing

The changelog will be updated on merging

The only problematic one is the JET tests. JET is a static analyzer, i.e. it attempts to detect bugs without running the code, i.e. logic bugs that would be pretty difficult to detect when running the code anyway because they might be in an edge case not covered by tests and rarely encountered by users. It does have false positives though, so I have simply upper bounded the number of errors it is permitted to detect. It was at 6, but now it detects 7 issues. Could you check if any of these issues were caused by something changed in this PR. I copy them here for convenience:

```
 rep = ═════ 7 possible errors found ═════
┌ commutator(o1::SymbolicUtils.Symbolic{…}, o2::SymbolicUtils.Symbolic{…}) @ QuantumSymbolics /home/runner/work/QuantumSymbolics.jl/QuantumSymbolics.jl/src/QSymbolicsBase/basic_ops_homogeneous.jl:209
│┌ QuantumSymbolics.SCommutator(o1::SymbolicUtils.Symbolic{…}, o2::SymbolicUtils.Symbolic{…}) @ QuantumSymbolics /home/runner/work/QuantumSymbolics.jl/QuantumSymbolics.jl/src/QSymbolicsBase/basic_ops_homogeneous.jl:199
││┌ hcat(::SymbolicUtils.Symbolic{…}, ::SymbolicUtils.Symbolic{…}) @ Base ./abstractarray.jl:1944
│││┌ kwcall(::@NamedTuple{…}, ::typeof(cat), ::SymbolicUtils.Symbolic{…}, ::SymbolicUtils.Symbolic{…}) @ Base ./abstractarray.jl:1993
││││┌ cat(::SymbolicUtils.Symbolic{…}, ::SymbolicUtils.Symbolic{…}; dims::Val{…}) @ Base ./abstractarray.jl:1993
│││││┌ _cat(::Val{…}, ::SymbolicUtils.Symbolic{…}, ::SymbolicUtils.Symbolic{…}) @ Base ./abstractarray.jl:1796
││││││┌ _cat_t(::Val{…}, ::Type{…}, ::SymbolicUtils.Symbolic{…}, ::SymbolicUtils.Symbolic{…}) where T @ Base ./abstractarray.jl:1805
│││││││┌ __cat(::Matrix, ::Tuple{…}, ::Tuple{…}, ::SymbolicUtils.Symbolic{…}, ::SymbolicUtils.Symbolic{…}) @ Base ./abstractarray.jl:1812
││││││││┌ __cat_offset!(A::Matrix, shape::Tuple{…}, catdims::Tuple{…}, offsets::Tuple{…}, x::SymbolicUtils.Symbolic{…}, X::SymbolicUtils.Symbolic{…}) @ Base ./abstractarray.jl:1816
│││││││││┌ __cat_offset1!(A::Matrix, shape::Tuple{…}, catdims::Tuple{…}, offsets::Tuple{…}, x::SymbolicUtils.Symbolic{…}) @ Base ./abstractarray.jl:1825
││││││││││┌ _copy_or_fill!(A::Matrix, inds::Tuple{…}, x::SymbolicUtils.Symbolic{…}) @ Base ./abstractarray.jl:1832
│││││││││││┌ fill!(V::SubArray{Bool, 2, Union{}, Tuple{…}, false}, x::SymbolicUtils.Symbolic{QuantumInterface.AbstractOperator}) @ Base ./multidimensional.jl:1512
││││││││││││┌ getproperty(x::SubArray{Bool, 2, Union{}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, f::Symbol) @ Base ./Base.jl:37
│││││││││││││ invalid builtin function call: Base.getfield(x::SubArray{Bool, 2, Union{}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false}, f::Symbol)
││││││││││││└────────────────────
│││││││││││┌ fill!(V::SubArray{…}, x::SymbolicUtils.Symbolic{…}) where Tv @ SparseArrays /cache/build/builder-amdci4-0/julialang/julia-release-1-dot-10/usr/share/julia/stdlib/v1.10/SparseArrays/src/sparsematrix.jl:3168
││││││││││││┌ getproperty(x::SubArray{Tv, 2, Union{}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false} where Tv, f::Symbol) @ Base ./Base.jl:37
│││││││││││││ invalid builtin function call: Base.getfield(x::SubArray{Tv, 2, Union{}, Tuple{UnitRange{Int64}, UnitRange{Int64}}, false} where Tv, f::Symbol)
││││││││││││└────────────────────
┌  @ QuantumSymbolics /home/runner/work/QuantumSymbolics.jl/QuantumSymbolics.jl/src/QSymbolicsBase/predefined.jl:314
│ no matching method found `QuantumSymbolics.SMulOperator(::QuantumSymbolics.SInvOperator, ::QuantumSymbolics.SOperator)`: SMulOperator(invop::QuantumSymbolics.SInvOperator, op::QuantumSymbolics.SOperator)
└────────────────────
┌  @ QuantumSymbolics /home/runner/work/QuantumSymbolics.jl/QuantumSymbolics.jl/src/QSymbolicsBase/predefined.jl:315
│ no matching method found `QuantumSymbolics.SMulOperator(::QuantumSymbolics.SOperator, ::QuantumSymbolics.SInvOperator)`: SMulOperator(op::QuantumSymbolics.SOperator, invop::QuantumSymbolics.SInvOperator)
└────────────────────
┌ (::QuantumSymbolics.var"#43#45")(__MATCHES__::Any) @ QuantumSymbolics /home/runner/.julia/packages/SymbolicUtils/0opve/src/rule.jl:314
│ `QuantumSymbolics.xs` is not defined: QuantumSymbolics.prefactorscalings_rule(QuantumSymbolics.xs)
└────────────────────
┌ apply_recipe(x::QuantumSymbolics.SDagger; kwargs...) @ QuantumSymbolics /home/runner/work/QuantumSymbolics.jl/QuantumSymbolics.jl/src/QSymbolicsBase/latexify.jl:31
│┌ getproperty(x::QuantumSymbolics.SDagger, f::Symbol) @ Base ./Base.jl:37
││ type SDagger has no field ket: Base.getfield(x::QuantumSymbolics.SDagger, f::Symbol)
│└────────────────────
┌ basis(x::QuantumSymbolics.StabilizerState) @ QuantumSymbolics /home/runner/work/QuantumSymbolics.jl/QuantumSymbolics.jl/src/extensions.jl:25
│┌ ^(b::SpinBasis{1//2, Int64}, N::Integer) @ QuantumInterface /home/runner/.julia/packages/QuantumInterface/VISSI/src/bases.jl:100
││┌ collect(itr::Base.Generator{_A, QuantumInterface.var"#3#4"{SpinBasis{1//2, Int64}}} where _A) @ Base ./array.jl:844
│││┌ collect_to_with_first!(dest::Any, v1::SpinBasis{…}, itr::Base.Generator{…} where _A, st::Any) @ Base ./array.jl:875
││││┌ grow_to!(dest::AbstractDict{…}, itr::Base.Generator{…} where _A, st::Any) where {…} @ Base ./dict.jl:133
│││││┌ indexed_iterate(I::SpinBasis{1//2, Int64}, i::Int64) @ Base ./tuple.jl:95
││││││ no matching method found `iterate(::SpinBasis{1//2, Int64})`: x = iterate(I::SpinBasis{1//2, Int64})
│││││└────────────────────
```

Krastanov:
Looks good, feel free to merge after you add a changelog and all tests except "performance tracking" and "codecov project" pass. Some more details on merging etiquette were posted in the other PR

PR #51, opened 2024-06-21
Defining literal quantum objects with macros

apkille:
Enables users to quickly define symbolic bras, kets, and operators with macros. If a basis for the quantum object is not specified, then by default the spin-1/2 basis will be selected.
```
julia> @sket k
|k⟩

julia> @sbra b
⟨b|

julia> @sop A
A

julia> b*k
⟨b||k⟩

julia> A*k
A|k⟩

julia> QuantumSymbolics.basis(k)
Spin(1/2)

julia> @sbra b FockBasis(2)
⟨b|

julia> QuantumSymbolics.basis(b)
Fock(cutoff=2)
```


Krastanov:
after the JET tests pass feel free to merge. A couple of guidelines:

- unless you have a specific reason, always use "Squash and merge", not rebase or merge commit options from the github interface
- before merge, please make an extra commit to update the CHANGELOG.
- if you plan to immediately make a new version release update the Project.toml to `0.3.1` and use today's date in the CHANGELOG header for 0.3.1
- if you plan to wait a bit to merge other things (personal choice, but generally not necessary to wait) update the Project.toml to `0.3.1-dev` and use the string `dev` instead of a date in the CHANGELOG header for 0.3.1
- if you want to make a public release, after this is merged go to the page for the very latest commit in this repository and leave a comment `@JuliaRegistrator register`

Krastanov:
welp... I JuliaRegistrator was a bit too trigger happy :D  I hoped it will not respond if I put its handle in backticks.

Krastanov:
as mentioned on the other pull request: actually, it seems the JET tests fail on master too, because of something I must have merged last minute as I was cleaning up the tests. You can just bump the threshold to 7 and not bother with them

Issue #54, opened 2024-06-24
Wrong distributive scaling properties 


apkille:
```
julia> @op A
A

julia> (1+im)*A
1 + 1imA
```

apkille:
Seems to be that `countmap_flatten` doesn't support multiplication of complex objects by general complex numbers. Will address in upcoming PR.

apkille:
Also, we currently have
```
julia> @bra b; @ket k;

julia> (b*k)*A
ERROR: TypeError: non-boolean (SymbolicUtils.BasicSymbolic{Bool}) used in boolean context
Stacktrace:
 [1] *(c::SBraKet, x::SOperator)
   @ QuantumSymbolics ~/Documents/Julia Packages/QuantumJulia/QuantumSymbolics.jl/src/QSymbolicsBase/basic_ops_homogeneous.jl:33
 [2] top-level scope
   @ REPL[16]:1
```

apkille:
> Could be that printing functionalities are misleading here:
> 
> ```
> julia> QuantumSymbolics.arguments((1+im)*A)
> 2-element Vector{Any}:
>  1 + 1im
>     A
> ```
> 
> Will address in upcoming PR.

Another printing issue is when you multiply a tensor product with an operator, say `A*(B ⊗ C)', which currently gives `AB⊗C`, rather than the preferred `A(B ⊗ C)`.

Issue #53, opened 2024-06-24
Unable to take tensor product of outer products


apkille:
```
julia> @ket k; @bra b;

julia> (k*b) ⊗ (k*b)
ERROR: MethodError: *(::SKet, ::SKet) is ambiguous.

Candidates:
  *(x::SymbolicUtils.Symbolic{T}, c) where T<:Union{QuantumInterface.AbstractBra, QuantumInterface.AbstractKet, QuantumInterface.AbstractOperator, QuantumInterface.AbstractSuperOperator}
    @ QuantumSymbolics ~/Documents/Julia Packages/QuantumJulia/QuantumSymbolics.jl/src/QSymbolicsBase/basic_ops_homogeneous.jl:34
  *(c, x::SymbolicUtils.Symbolic{T}) where T<:Union{QuantumInterface.AbstractBra, QuantumInterface.AbstractKet, QuantumInterface.AbstractOperator, QuantumInterface.AbstractSuperOperator}
    @ QuantumSymbolics ~/Documents/Julia Packages/QuantumJulia/QuantumSymbolics.jl/src/QSymbolicsBase/basic_ops_homogeneous.jl:33

Possible fix, define
  *(::SymbolicUtils.Symbolic{T}, ::SymbolicUtils.Symbolic{T}) where {T<:Union{…}, T<:Union{…}}

Stacktrace:
 [1] prefactorscalings(xs::Vector{SOuterKetBra}; scalar::Bool)
   @ QuantumSymbolics ~/Documents/Julia Packages/QuantumJulia/QuantumSymbolics.jl/src/QSymbolicsBase/rules.jl:26
 [2] prefactorscalings
   @ ~/Documents/Julia Packages/QuantumJulia/QuantumSymbolics.jl/src/QSymbolicsBase/rules.jl:19 [inlined]
 [3] STensorOperator(terms::Vector{SOuterKetBra})
   @ QuantumSymbolics ~/Documents/Julia Packages/QuantumJulia/QuantumSymbolics.jl/src/QSymbolicsBase/basic_ops_homogeneous.jl:158
 [4] tensor(::SOuterKetBra, ::SOuterKetBra)
   @ QuantumSymbolics ~/Documents/Julia Packages/QuantumJulia/QuantumSymbolics.jl/src/QSymbolicsBase/basic_ops_homogeneous.jl:170
 [5] top-level scope
   @ REPL[4]:1
[...]

PR #55, opened 2024-06-26
trace, partial trace, and superoperators

apkille:
Added trace and partial trace functionalities with examples and tests. Also created a symbolic superoperator `SSuperOperator`, including an option to express instances of this composite type in the Kraus representation with the function `kraus`. Future work can be done to add additional superoperator representations and conversions between each other.

In the process of developing this branch, I encountered a few bugs mentioned in #53 and #54, which I fix in this PR.


apkille:
The failed tests seem to come from the recent bumped compat for TermInterface #52. 
```
 MethodError: no method matching sorted_arguments(::QuantumSymbolics.STrace)
│    
│    Closest candidates are:
│      sorted_arguments(::SymbolicUtils.PolyForm)
│       @ SymbolicUtils ~/.julia/packages/SymbolicUtils/dtCid/src/polyform.jl:235
│      sorted_arguments(::SymbolicUtils.BasicSymbolic)
│       @ SymbolicUtils ~/.julia/packages/SymbolicUtils/dtCid/src/types.jl:119
```

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -164,21 +164,21 @@ iscall(::STensor) = true
 arguments(x::STensor) = x.terms
 operation(x::STensor) = ⊗
 head(x::STensor) = :⊗
-children(x::STensor) = pushfirst!(x.terms,:⊗)
```
Krastanov, line 167: whoops, we missed this in a previous review

src/QSymbolicsBase/predefined.jl
```diff
@@ -286,7 +286,109 @@ function Base.show(io::IO, x::SDagger)
...
+head(x::SPartialTrace) = :ptrace
+children(x::SPartialTrace) = [:ptrace, x.obj, x.sys]
+Base.show(io::IO, x::SPartialTrace) = print(io, "tr$(x.sys)($(x.obj))")
+ptrace(x::Symbolic{AbstractOperator}, s) = SPartialTrace(x, s)
+function ptrace(x::STensorOperator, s)
+    terms = arguments(x)
+    sys_op = terms[s]
+    new_terms = deleteat!(copy(terms), s)
+    isone(length(new_terms)) ? tr(sys_op)*first(new_terms) : tr(sys_op)*STensorOperator(new_terms)
+end
```
Krastanov, line 376: This is actually a bit worse than this. You can have `@op A SpinBasis(1//2)^2`, i.e. the symbolic operator is already over a composite basis. The number of subsystem would be the number of subsystems in `basis`, which might be different from the number of terms in the tensor product.

Also, what if flattening has not happened. E.g `ptrace(A⊗(B⊗C+D⊗E), 2)`?

Krastanov (changes requested):
great start! There are a few things I want us to discuss before considering for merging

Krastanov:
Just so that I do not get lost, could you point out in this comment thread what are the fixes for #53 and #54? And the corresponding test?

Krastanov:
For the spellchecker issue, `BA` can be whitelisted in https://github.com/QuantumSavory/QuantumSymbolics.jl/blob/main/.typos.toml

src/QSymbolicsBase/literal_objects.jl
```diff
@@ -92,4 +106,4 @@ symbollabel(x::SZero) =  "𝟎"
 basis(x::SZero) = nothing
 
 Base.show(io::IO, x::SZero) = print(io, symbollabel(x))
-Base.iszero(x::SymQObj) = isa(x, SZero)
+Base.iszero(x::Union{SymQObj, Symbolic{Number}, Symbolic{Complex}}) = isa(x, SZero)
```
Krastanov, line 109: I think the new `iszero` implementations here are type piracy. Where are they needed?

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -103,11 +108,6 @@ function Base.show(io::IO, x::SAddOperator)
     ordered_terms = sort([repr(i) for i in arguments(x)])
     print(io, "("*join(ordered_terms,"+")::String*")") # type assert to help inference
 end
-const SAddBra = SAdd{AbstractBra}
-function Base.show(io::IO, x::SAddBra)
-    ordered_terms = sort([repr(i) for i in arguments(x)])
-    print(io, "("*join(ordered_terms,"+")::String*")") # type assert to help inference
-end
```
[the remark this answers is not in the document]
apkille: It was only because in other parts of the code I would typically order things bra, ket, operator. I can undo it if you would like.

src/QSymbolicsBase/predefined.jl
apkille: Ah, these are both fair comments. Will make changes.

src/QSymbolicsBase/basic_ops_homogeneous.jl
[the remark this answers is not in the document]
Krastanov: nah, sounds good, no need to change it again

src/QSymbolicsBase/QSymbolicsBase.jl
```diff
@@ -171,10 +173,10 @@ propsequal(x,y) = all(n->isequal(getproperty(x,n),getproperty(y,n)), propertynam
 ##
 
 include("literal_objects.jl")
+include("predefined_CPTP.jl")
 include("basic_ops_homogeneous.jl")
 include("basic_ops_inhomogeneous.jl")
 include("predefined.jl")
-include("predefined_CPTP.jl")
```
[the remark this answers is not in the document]
Krastanov: In my mind "predefined" is meant for some well known specific instances of symbolic types, like the number pi, or the Pauli operator X, not for types that can encompass many instances. Maybe it would make sense to move some of the stuff out from this file and into a new `basic_superops.jl`?

src/QSymbolicsBase/rules.jl
```diff
@@ -22,7 +22,7 @@ function prefactorscalings(xs; scalar=false) # If the scalar keyword is true, th
     for x in xs
         if isexpr(x) && operation(x) == *
             c,t = arguments(x)
-            if !scalar
+            if !scalar && !isa(x,SOuterKetBra)
```
apkille, line 25: Fix for #53.

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -45,7 +45,7 @@ function Base.show(io::IO, x::SScaledKet)
 end
 const SScaledOperator = SScaled{AbstractOperator}
 function Base.show(io::IO, x::SScaledOperator)
-    if x.coeff isa Number
+    if x.coeff isa Real
```
apkille, line 48: Fix for #54.

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -164,21 +164,21 @@ iscall(::STensor) = true
...
 end
 basis(x::STensor) = tensor(basis.(x.terms)...)
 
+const STensorBra = STensor{AbstractBra}
+Base.show(io::IO, x::STensorBra) = print(io, join(map(string, arguments(x)),""))
 const STensorKet = STensor{AbstractKet}
 Base.show(io::IO, x::STensorKet) = print(io, join(map(string, arguments(x)),""))
 const STensorOperator = STensor{AbstractOperator}
-Base.show(io::IO, x::STensorOperator) = print(io, join(map(string, arguments(x)),"⊗"))
+Base.show(io::IO, x::STensorOperator) = print(io, "("*join(map(string, arguments(x)),"⊗")*")")
```
apkille, line 179: Another fix in comment of #54.

src/QSymbolicsBase/literal_objects.jl
apkille: This fixes a bug mentioned in #54 as well: https://github.com/QuantumSavory/QuantumSymbolics.jl/issues/54#issuecomment-2187334722.

src/QSymbolicsBase/basic_ops_homogeneous.jl
apkille: https://github.com/QuantumSavory/QuantumSymbolics.jl/issues/54#issuecomment-2187330143

src/QSymbolicsBase/QSymbolicsBase.jl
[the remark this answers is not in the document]
apkille: I was thinking about this too and agree. I think we should also move linear algebra operations like `dagger` and `trace` out of `predefined.jl` into a new file, perhaps 'linalg.jl`?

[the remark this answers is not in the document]
Krastanov: good idea. I like the shorter name.

src/QSymbolicsBase/predefined.jl
apkille: Just found another small bug:
```
julia> @op A; @op B; @op C;

julia> tp = A⊗B⊗C
((A⊗B)⊗C)

julia> QuantumSymbolics.arguments(tp)
2-element Vector{Any}:
 (A⊗B)
 C
```
We would want `tp` to show `(A⊗B⊗C)` and `QuantumSymbolics.arguments(tp)` to be a 3-element vector of the symbolic operators.

Weird. I wonder why this doesn't happen for multiplication of operators:
```
julia> mp = A*B*C
ABC

julia> QuantumSymbolics.arguments(mp)
3-element Vector{Any}:
 A
 B
 C
```

apkille: > Also, what if flattening has not happened. E.g `ptrace(A⊗(B⊗C+D⊗E), 2)`?
 
A question somewhat related to this: should we have a function in the package that expands expressions? For instance, if a user for some reason wanted to convert `A⊗(B⊗C+D⊗E)` into `A⊗B⊗C+A⊗D⊗E`, shouldn't they have a way to do that? Maybe we could call it `qexpand`, using the `q` prefix the same we do for `qsimplify`.

Krastanov: yup, let's have a `qexpand` for now. Maybe at some point we will just hook into the standard `expand` and at that point it can be just an alias without making a breaking release

concerning the difference between mul and tensor, I think it is because `prefactorscalings` were created for tensor, and it has been kinda hackishly applied to mul. We probably need to clean `prefactorscalings` to work on both. I will make a separate issue for that.

Krastanov: I created but then deleted the issue because I had misunderstood something when I was writing it. Could you remind me why you added `scalars=false` options to `prefactorscalings`? I do not think I understand it. I see it has something to do with `SCommutator` but I do not get it.

Krastanov: To finally actually answer your question about why * and ⊗ behave differently in terms of flattening, it is because of the Julia parser. Julia parses repeated * as a single multi-argument expression, while it parses ⊗ as a nested expression. The author of the parser has expressed uncertainty whether that was a good decision. One reason it can in principle be useful, is because it permits optimizing the order of matrix multiplications. But maybe it should not have been special cased, rather done for all infix operators.

```
julia> :(a*b*c) |> dump
Expr
  head: Symbol call
  args: Array{Any}((4,))
    1: Symbol *
    2: Symbol a
    3: Symbol b
    4: Symbol c

julia> :(a⊗b⊗c) |> dump
Expr
  head: Symbol call
  args: Array{Any}((3,))
    1: Symbol ⊗
    2: Expr
      head: Symbol call
      args: Array{Any}((3,))
        1: Symbol ⊗
        2: Symbol a
        3: Symbol b
    3: Symbol c

```

Krastanov: besides simplifying and hopefully removing the `scalar=false` from prefactorscaling, maybe we should add a simple `flattenop` function that does

`flattenop(*, [A, B*C, D, E⊗F, G+H])` = [A, B, C, D, E⊗F, G+H]
`flattenop(+, [A, B*C, D, E⊗F, G+H])` = [A, B*C, D, E⊗F, G, H]
`flattenop(⊗, [A, B*C, D, E⊗F, G+H])` = [A, B*C, D, E, F, G+H]

and pretty much always use them together

apkille: I rewrote `scalar=false` options to `prefactorscalings` so that I could pull only a scalar out from a subexpression rather than, say an entire operator. This was important with expressions such as commutators, where if I had the commutator of a scalar operator, I could factor out the scalar value. For instance, `commutator(2*A, B)` would give `2*commutator(A, B)`. In retrospect, this is a messy fix, so that's my bad. I will dedicate time to cleaning up the flattening procedures.

Krastanov: Could you give an example where the commutator would misbehave if the `scalar` keyword did not exist?

apkille: If the `scalar` keyword did not exist, then for example, `commutator(A*B, C)` would give `A*[B,C]`, which is different than the correct property `[A,C]*B + A*[B,C]`.

Krastanov: what if in `commutator` we do not do `prefactorscalings([op1, op2])` but rather do:

```
c1, t1 = prefactorscalings(op1)
head1 = length(t1)==1 ? identity : operation(op1)
c2, t2...
c1*c2 * new(head1(c1...), head2(c2...))
```

apkille: Would it be good to create another file that handles utilities like `flattenop` and `prefactorscalings`, perhaps `utils.jl`? As it doesn't make much sense to have them in `rules.jl` where we are defining simplification rules.

Krastanov: Wait, no that is overly complicated. Why not just have `scalar=true` always. Is there a situation that benefits from `scalar=false`?

apkille: > what if in `commutator` we do not do `prefactorscalings([op1, op2])` but rather do:
> 
> ```
> c1, t1 = prefactorscalings(op1)
> head1 = length(t1)==1 ? identity : operation(op1)
> c2, t2...
> c1*c2 * new(head1(c1...), head2(c2...))
> ```
Yes that sounds good.

Krastanov: yes about a new file

apkille: It's my understanding that `prefactorscalings` was originally created for tensor product properties. For instance, `(A|k1⟩)⊗|k2⟩` would give `A*(k1⟩⊗|k2⟩)`.

apkille: Regardless, I will clean up the scaling and flattening functions.

Krastanov: `prefactorscaling` was always meant only for scalars. The operator identity you are giving is wrong because A is defined to act on the Hilbert space of k2, not on the composite hilbert space of k1k2.

apkille:
I marked this PR as a draft because I first want to clean up the scaling and flattening functions, and also implement a `qexpand` function for expanding quantum expressions. Since this PR adds a lot of features, I would like to submit these changes in a separate PR.

src/QSymbolicsBase/basic_superops.jl
```diff
@@ -0,0 +1,64 @@
...
+    sop
+    krausops
+end
+isexpr(::KrausRepr) = true
+iscall(::KrausRepr) = true
+arguments(x::KrausRepr) = [x.sop, x.krausops]
+operation(x::KrausRepr) = kraus
+head(x::KrausRepr) = :kraus
+children(x::KrausRepr) = [:kraus, x.sop, x.krausops]
+kraus(s::Symbolic{AbstractSuperOperator}, xs::Symbolic{AbstractOperator}...) = KrausRepr(s,collect(xs))
```
Krastanov, line 30: I do not think this matches the `arguments` and `children` definitions given above. But if the definitions above are changed as I suggested in the other comment, this should be fine.

src/QSymbolicsBase/basic_superops.jl
```diff
@@ -0,0 +1,64 @@
...
+symbollabel(x::KrausRepr) = symbollabel(x.sop)
```
Krastanov, line 31: I guess if `sop` is dropped, this will be removed as well.

src/QSymbolicsBase/basic_superops.jl
```diff
@@ -0,0 +1,64 @@
...
+basis(x::KrausRepr) = basis(x.sop)
+Base.show(io::IO, x::KrausRepr) = print(io, symbollabel(x))
```
Krastanov, line 33: and this will need to be changed, maybe to something like `𝒦[op1, op2, op3, ...]` or something else if there is a popular notation for Kraus maps.

src/QSymbolicsBase/linalg.jl
```diff
@@ -89,6 +89,140 @@ function Base.show(io::IO, x::SDagger)
...
+    tensor(new_bases...)
+end
+Base.show(io::IO, x::SPartialTrace) = print(io, "tr$(x.sys)($(x.obj))")
+function ptrace(x::Symbolic{AbstractOperator}, s) 
+    ex = isexpr(x) ? qexpand(x) : x
+    if isa(ex, typeof(x))
+        if isa(basis(x), CompositeBasis)
+            SPartialTrace(x, s)
+        else
+            throw(ArgumentError("cannot take partial trace of a single quantum system"))
```
Krastanov, line 188: If `s==1` this would be just a trace. Otherwise, yeah, it should throw for `!=1`.

src/QSymbolicsBase/linalg.jl
```diff
@@ -89,6 +89,140 @@ function Base.show(io::IO, x::SDagger)
...
+end
+function ptrace(x::STensorOperator, s)
+    ex = qexpand(x)
+    if isa(ex, SAddOperator)
+        ptrace(ex, s)
+    else
+        terms = arguments(ex)
+        newterms = []
+        if isa(basis(terms[s]), CompositeBasis)
+            SPartial(ex, s)
```
Krastanov, line 217: `SPartial`? I guess this is a typo. But if the typo was not caught by tests, we should add tests that would have caught it.

Krastanov:
The downstream buildkite error is known and unrelated.

src/QSymbolicsBase/linalg.jl
```diff
@@ -89,6 +89,140 @@ function Base.show(io::IO, x::SDagger)
...
+        else 
+            sys_op = terms[s]
+            new_terms = deleteat!(copy(terms), s)
+            isone(length(new_terms)) ? tr(sys_op)*first(new_terms) : tr(sys_op)*STensorOperator(new_terms)
+        end
+    end
+end
```
Krastanov, line 224: I have not looked through these very carefully yet, but one thing to ensure with more complicated code like this is that it has 100% test coverage. Codecov has made an automatic comment that links to this report https://app.codecov.io/gh/QuantumSavory/QuantumSymbolics.jl/pull/55?src=pr&el=tree&filepath=src%2FQSymbolicsBase%2Flinalg.jl#diff-c3JjL1FTeW1ib2xpY3NCYXNlL2xpbmFsZy5qbA

You can use `@test_throw` to check that errors are raised when expected.

test/test_expand.jl
```diff
@@ -26,6 +26,9 @@ using Test
     @test isequal(qexpand((B+C+D)*A), B*A + C*A + D*A)
     @test isequal(qexpand(commutator(A, B) * C), A*B*C - B*A*C)
 
+    @test isequal(qexpand(A*(B⊗C⊗D)), (A*B)⊗(A*C)⊗(A*D))
+    @test isequal(qexpand((B⊗C⊗D)*A), (B*A)⊗(C*A)⊗(D*A))
```
Krastanov, line 30: These two seem wrong.

For `A*(B⊗C⊗D)` one must have `basis(A)==basis(B⊗C⊗D)`.

For `(A*B)⊗(A*C)⊗(A*D)` one must have `basis(A)==basis(B)==basis(C)==basis(D)`.

We should probably have a basis consistency check in the operator multiplication function (and we can add such checks in the apply functions as well).

test/test_expand.jl
```diff
@@ -26,6 +26,9 @@ using Test
...
+
     @test isequal(qexpand((A⊗B)*(C⊗D)), (A*C)⊗(B*D))
```
Krastanov, line 32: I did not notice this originally, but this is only true if `basis(A)==basis(C)` and `basis(B)==basis(D)`. Could you add a test that confirms an error is raised when this is not the case?

test/test_expand.jl
```diff
@@ -26,6 +26,9 @@ using Test
...
     @test isequal(qexpand((b₁⊗b₂)*(k₁⊗k₂)), (b₁*k₁)*(b₂*k₂))
```
Krastanov, line 33: same comment about this being valid only when there is an appropriate basis equality

test/test_trace.jl
```diff
@@ -0,0 +1,51 @@
...
+    @test isequal(ptrace(exp3, 3), tr(C)*(A⊗B) + tr(E)*(A⊗D))
+    @test isequal(QuantumSymbolics.basis(ptrace(exp3, 3)), SpinBasis(1//2)⊗SpinBasis(1//2))
+
+    @test isequal(ptrace(exp4, 1), tr(A*F)*((B*F)⊗(C*F)) + tr(A*F)*((D*F)⊗(E*F)))
+    @test isequal(QuantumSymbolics.basis(ptrace(exp4, 1)), SpinBasis(1//2)⊗SpinBasis(1//2))
+    @test isequal(ptrace(exp4, 2), tr(B*F)*((A*F)⊗(C*F)) + tr(D*F)*((A*F)⊗(E*F)))
+    @test isequal(QuantumSymbolics.basis(ptrace(exp4, 2)), SpinBasis(1//2)⊗SpinBasis(1//2))
+    @test isequal(ptrace(exp4, 3), tr(C*F)*((A*F)⊗(B*F)) + tr(E*F)*((A*F)⊗(D*F)))
+    @test isequal(QuantumSymbolics.basis(ptrace(exp4, 2)), SpinBasis(1//2)⊗SpinBasis(1//2))
+end
```
Krastanov, line 51: I have not had a chance to review this yet. I will leave it for the next round.

Krastanov (changes requested):
I did a quick review, I have not had a chance to go through everything yet. I think there are a few minor issues with basis checks.

src/QSymbolicsBase/QSymbolicsBase.jl
```diff
@@ -23,28 +23,30 @@ export SymQObj,QObj,
        apply!,
        express,
        tensor,⊗,
-       dagger,projector,commutator,anticommutator,
+       dagger,projector,commutator,anticommutator,expand,tr,ptrace,
```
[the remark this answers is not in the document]
apkille: No, good catch, I must've missed this when I was merging.

src/QSymbolicsBase/basic_superops.jl
```diff
@@ -0,0 +1,64 @@
...
+julia> K = kraus(ℰ, A₁, A₂, A₃);
+
+julia> @op ρ;
+
+julia> K*ρ
+(A₁ρA₁†+A₂ρA₂†+A₃ρA₃†)
+```
+"""
+@withmetadata struct KrausRepr <: Symbolic{AbstractSuperOperator}
+    sop
```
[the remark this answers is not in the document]
apkille: I agree with you, we wouldn't want `KrausRepr` to wrap a symbolic superoperator.

src/QSymbolicsBase/linalg.jl
apkille: That's my bad, I will make additional tests.

test/test_expand.jl
apkille: I was thinking about this issue when I was writing docs earlier today. I suppose we would also want to have a consistency check in our definition for `isequal`?

Krastanov:
Just FYI, QuantumInterface.jl provides some basis checking foundations and error types: https://github.com/qojulia/QuantumInterface.jl/blob/main/src/bases.jl#L140-L180

src/QSymbolicsBase/linalg.jl
```diff
@@ -89,6 +89,160 @@ function Base.show(io::IO, x::SDagger)
...
+                isone(length(new_terms)) ? push!(add_terms, trace*first(new_terms)) : push!(add_terms, trace*(⊗)(new_terms...))
+            end  
+        end
+        (+)(add_terms...)
+    elseif s==1
+        tr(x)
+    else
+        throw(ArgumentError("cannot take partial trace of a single quantum system"))
+    end
+end
```
apkille, line 228: I should explain my thinking behind this complicated method, as it would be good for me to have my thoughts organized in writing and it would be helpful for review.

 It is intended to take the partial trace of a summation of operators. First, we must check if the basis of the summation is a composite basis, and if not, then the partial trace must be over the first system (in that case we would just take a regular trace of the sum). If neither of these conditions are true, then we throw an argument error, as then we would be taking the partial trace of a single quantum system.

If the basis of the summation is a composite basis, then we iterate through the arguments of the sum. Here is where it gets complicated. A term in the arguments could be (1) a tensor product of operators with non-composite bases, (2) a tensor product of operators with composite bases, (3) an operator with composite basis (could be `SMulOperator` or just a single operator, or (4) a scaled version of (1-3). That is why we have that first if-else statement in the for loop.

Then, we have the `if any(j ->...` statement to deal with the case where we have a tensor product of operators with composite bases, in which case we can't take the trace of any of the operators, so we just return `SPartialTrace`. If that condition is not satisfied, i.e., we have a summation of tensor products of operators, each with non-composite bases, then we proceed with tracing over the specified system.

Hopefully this helps. I'm happy to clarify anything.

test/test_basis_consistency.jl
```diff
@@ -11,3 +13,14 @@ using QuantumSymbolics
...
+@op A SpinBasis(1//2) ⊗ SpinBasis(1//2); @op B; 
+@ket k; @bra b; @ket l SpinBasis(1//2) ⊗ SpinBasis(1//2);
+
+@test_throws IncompatibleBases A*B
+@test_throws IncompatibleBases commutator(A, B)
+@test_throws IncompatibleBases anticommutator(A, B)
+@test_throws IncompatibleBases A*k
+@test_throws IncompatibleBases b*A
+@test_throws IncompatibleBases l*b
+@test_throws IncompatibleBases b*l
```
Krastanov, line 26: could you make a file just for trows? E.g. `test_throws`?

Krastanov (changes requested):
looks good!

I left a bunch of comments related to an annoyance with QuantumInterface and created an issue to track it https://github.com/qojulia/QuantumInterface.jl/issues/25

I would suggest merging this **without** addressing that issue, as otherwise there will be too much delay (and it is not a correctness issue).

I left a few other minor comments. Feel free to merge this when you have chance to address them.

Thanks for setting all of this up!

src/QSymbolicsBase/basic_superops.jl
```diff
@@ -0,0 +1,61 @@
...
+    krausops
+end
+isexpr(::KrausRepr) = true
+iscall(::KrausRepr) = true
+arguments(x::KrausRepr) = x.krausops
+operation(x::KrausRepr) = kraus
+head(x::KrausRepr) = :kraus
+children(x::KrausRepr) = [:kraus; x.krausops]
+kraus(xs::Symbolic{AbstractOperator}...) = KrausRepr(collect(xs))
+basis(x::KrausRepr) = basis(first(x.krausops))
```
Krastanov, line 29: coverage review: could you add a test for this line

src/QSymbolicsBase/basic_superops.jl
```diff
@@ -0,0 +1,61 @@
...
+# Superoperator operations
+##
+
+"""Symbolic application of a superoperator on an operator
+
+```jldoctest
+julia> @op A; @superop S;
+
+julia> S*A
+S[A]
```
Krastanov, line 42: this doctest is not closed so it is not running

src/QSymbolicsBase/basic_superops.jl
```diff
@@ -0,0 +1,61 @@
...
+operation(x::SSuperOpApply) = *
+head(x::SSuperOpApply) = :*
+children(x::SSuperOpApply) = [:*,x.sop,x.op]
+Base.:(*)(sop::Symbolic{AbstractSuperOperator}, op::Symbolic{AbstractOperator}) = SSuperOpApply(sop,op)
+Base.:(*)(sop::Symbolic{AbstractSuperOperator}, op::SZeroOperator) = SZeroOperator()
+Base.:(*)(sop::Symbolic{AbstractSuperOperator}, k::Symbolic{AbstractKet}) = SSuperOpApply(sop,SProjector(k))
+Base.:(*)(sop::Symbolic{AbstractSuperOperator}, k::SZeroKet) = SZeroOperator()
+Base.:(*)(sop::KrausRepr, op::Symbolic{AbstractOperator}) = (+)((i*op*dagger(i) for i in sop.krausops)...)
+Base.:(*)(sop::KrausRepr, k::Symbolic{AbstractKet}) = (+)((i*SProjector(k)*dagger(i) for i in sop.krausops)...)
+Base.show(io::IO, x::SSuperOpApply) = print(io, "$(x.sop)[$(x.op)]")
```
Krastanov, line 60: coverage review: could you add a test for this line

This line being reported as uncovered, is how I noticed that the doctest above does not work.

src/QSymbolicsBase/literal_objects.jl
```diff
@@ -67,15 +67,29 @@ SHermitianUnitaryOperator(name) = SHermitianUnitaryOperator(name, qubit_basis)
...
+end
+SSuperOperator(name) = SSuperOperator(name, qubit_basis)
+macro superop(name, basis)
+    :($(esc(name)) = SSuperOperator($(Expr(:quote, name)), $(basis)))
+end
+macro superop(name)
+    :($(esc(name)) = SSuperOperator($(Expr(:quote, name))))
+end
+ishermitian(x::SSuperOperator) = false
+isunitary(x::SSuperOperator) = false
```
Krastanov, line 82: coverage review: could you add a test for these two lines

Krastanov (dismissed):
I forgot to check the coverage. There are a couple of lines that would be nice to test (and one of them is related to a broken doctest)

src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -139,15 +139,19 @@ children(x::SMulOperator) = [:*;x.terms]
 function Base.:(*)(xs::Symbolic{AbstractOperator}...) 
     zero_ind = findfirst(x->iszero(x), xs)
     if isnothing(zero_ind)
-        terms = flattenop(*, collect(xs))
-        coeff, cleanterms = prefactorscalings(terms)
-        coeff * SMulOperator(cleanterms)
+        if any(x->!(samebases(basis(x),basis(first(xs)))),xs)
```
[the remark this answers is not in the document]
apkille: I agree. Next week I'll make a PR to QuantumInterface and then clean this up a bit.

src/QSymbolicsBase/linalg.jl
apkille: Ah yeah, maybe we could use rewriters from Symbolics for this. Although they can get messy too, especially when you have multiple predicates involved. When I was making this, I was mostly focused on capturing all of the different cases and making it easy to follow (even if it is indeed long). I'll think about this more but for now I say we leave it as is.

Issue #49, opened 2024-06-19
latexify partially broken after #40


Krastanov:
latexify(dagger(X1)) does not work anymore after #40 due to the changes to `dagger`

detected by JET (those tests need some polish to be more reliable)

@apkille

PR #57, opened 2024-06-28
express, qsimplify, and latexify updates

apkille:
- Added a doc that describes features of `express` and added more tests in `test_express_cliff.jl`.
- `qsimplify` can now traverse through subexpressions using `Prewalk` from `SymbolicUtils.jl`. This required defining `maketerm` from `TermInterface.jl` for each symbolic operation.
- Added latexify capabilities and fixed bug mentioned in #49 .


src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -30,30 +30,39 @@ arguments(x::SScaled) = [x.coeff,x.obj]
...
+    else 
+        x isa SScaled ? SScaled{T}(c*x.coeff, x.obj) : SScaled{T}(c, x) 
+    end
+end
 Base.:(*)(x::Symbolic{T}, c) where {T<:QObj} = c*x
 Base.:(/)(x::Symbolic{T}, c) where {T<:QObj} = iszero(c) ? throw(DomainError(c,"cannot divide QSymbolics expressions by zero")) : (1/c)*x
 basis(x::SScaled) = basis(x.obj)
 
 const SScaledKet = SScaled{AbstractKet}
+maketerm(::Type{SScaledKet}, f, a, t, m) = f(a...)
```
Krastanov, line 45: I think it is my bad on missing this new requirement of the API.

But is this using it correctly? In the docs I see it as 4-arg function, while here there are 5 args: https://juliasymbolics.github.io/TermInterface.jl/dev/

Krastanov: also, if the vast majority are the same call can't we just write a single method along the lines of

```
maketerm(::Type{<:Symbolic{<:QObj}}, ...) = ...
```

docs/src/express.md
```diff
@@ -0,0 +1,80 @@
...
+julia> ψ.metadata
+QuantumSymbolics.Metadata(Dict{Tuple{AbstractRepresentation, AbstractUse}, Any}((CliffordRepr(), UseAsState()) => MixedDestablizer 1×1, (QuantumOpticsRepr(), UseAsState()) => Ket(dim=2)
+  basis: Spin(1/2)
+ 1.0 + 0.0im
+ 0.0 + 0.0im))
+```
+
+Here, we specified an instance of [`CliffordRepr`](@ref) in the second argument to convert `ψ` into a tableau of Pauli operators containing its stabilizer and destabilizer states. Now, both the state vector and Clifford representation of `ψ` have been cached.
+
+For Pauli operators, additional flexibility is given for translations to the Clifford formalism. Users have the option to convert a multi-qubit Pauli operator to an observable or operation with instances of [`UseAsObservable`](@ref) and [`UseAsOperation`](@ref), respectively. Take the Pauli operator $Y$, for example, which in `QuantumSymbolics` is the constants `Y` or `σʸ`:
```
Krastanov, line 55: I would suggest a subheader just before this section with something like "Examples of Edge Cases" -- to note that these are not part of the big picture, but rather just details of implementation in some random corners.

Krastanov (dismissed):
Looks good! Feel free to merge after addressing the comments if tests pass.

src/QSymbolicsBase/basic_ops_homogeneous.jl
apkille: I don't think the docs for TermInterface are updated. The same can be said for most of the Symbolics libraries. It look some rummaging in the source code in SymbolicUtils to find out how Postwalk works. You can find it around the following line: https://github.com/JuliaSymbolics/SymbolicUtils.jl/blob/d009ca8e986b12982dce13794218ea6b632da617/src/rewriters.jl#L197.

apkille: By the way, if you haven't noticed already, don't use this website for examining SymbolicUtils documentation: https://symbolicutils.juliasymbolics.org. Even though it comes up as on of the first results, it's wildly out of date.

PR #69, opened 2024-07-12
Further development of Fock states

apkille:
- Adds a documentation page for using Fock states.
- Moved predefined Fock objects from `predefined.jl` to a new file `predefined_fock.jl`.
- Added displacement and phase shift operators, `DisplaceOp` and `PhaseShiftOp`, respectively.
- Added simplification rules and rewriter with tests for Fock states.
- Updated `express` functionality for Fock objects.


apkille:
@krastanov you might disagree with the following changes, which is OK and we can discuss them:
- I replaced `ContinuousCoherentState` and `DiscreteCoherentState` with simply `CoherentState`, as that distinction didn't seem too important to include in this package, particularly since QuantumOptics.jl only has `coherentstate`.
- In `express_nolookup`, before for Fock states we would change the dimension of the Fock basis to 2 with the warning that the cutoff wasn't defined, but of course this isn't always the case, as we can define a finite cutoff. So, if the cutoff is infinite, then we change the dimension to 2.

docs/src/fock.md
```diff
@@ -0,0 +1,147 @@
+# Fock States
+
+```@meta
+DocTestSetup = quote
+    using QuantumSymbolics, QuantumOptics
+end
+```
+
+In this section, we describe symbolic representations of Fock states in QuantumSymbolics, which can be numerically translated to `QuantumOptics.jl`. A Fock state is the number state representation $|n\rangle$ of a system with $n$ particles.
```
Krastanov, line 9: The definition of Fock state here is misleading. Something like "the number of excitation quanta of a single quantum harmonic oscillator" would be better. We have exactly one system here.

docs/src/fock.md
```diff
@@ -0,0 +1,147 @@
...
+
+One can define a basis of the Fock space with `FockBasis(N, offset=0)`, where `N` is the highest available Fock state and `offset` is the lowest cutoff state, which is by default zero. In the following example, we create a `FockState` with 3 quanta in an infinite-dimension Fock space:
+
+```jldoctest
+julia> b = FockBasis(Inf, 0.0)
+Fock(cutoff=Inf)
+
+julia> f = FockState(3, b)
+|3⟩
+```
```
Krastanov, line 19: Let's not tell people about `cutoff` in this tutorial. It is not well supported by the rest of the symbolic library and it is pretty niche.

docs/src/fock.md
```diff
@@ -0,0 +1,147 @@
+# Fock States
```
Krastanov, line 1: This page mentions in a couple of places things that are not Fock states. Maybe we can clarify this somewhere in the introductory paragraphs and change the title. Something like "Quantum Harmonic Oscillators" or "Bosonic Systems". And a description of how "we call the basis a `FockBasis` but we can symbolically represent arbitrary bosonic states in it".

docs/src/fock.md
```diff
@@ -0,0 +1,147 @@
...
+```jldoctest
+julia> b = FockBasis(Inf, 0.0);
+
+julia> f = FockState(3, b);
+
+julia> raise = Create*f
+a†|3⟩
+
+julia> qsimplify(raise, rewriter=qsimplify_fock)
+2.0|4⟩
```
Krastanov, line 48: The `2` should probably not be a float

docs/src/fock.md
```diff
@@ -0,0 +1,147 @@
...
+
+julia> lower = Destroy*f
+a|3⟩
+
+julia> qsimplify(lower, rewriter=qsimplify_fock)
+1.7320508075688772|2⟩
```
Krastanov, line 54: this should probably be a symbolic `sqrt(2)`

docs/src/fock.md
```diff
@@ -0,0 +1,147 @@
...
+D(im)
+
+julia> c = qsimplify(displace*vac, rewriter=qsimplify_fock)
+|im⟩
+
+julia> phase = PhaseShiftOp(pi, b)
+U(π)
+
+julia> qsimplify(phase*c, rewriter=qsimplify_fock)
+|1.2246467991473532e-16 - 1.0im⟩
```
Krastanov, line 93: Is there a symbolic pi in the Symbolics ecosystem that gives exact result here?

docs/src/fock.md
```diff
@@ -0,0 +1,147 @@
...
+
+```@meta
+DocTestSetup = quote
+    using QuantumSymbolics, QuantumOptics
+end
+```
+
+In this section, we describe symbolic representations of Fock states in QuantumSymbolics, which can be numerically translated to `QuantumOptics.jl`. A Fock state is the number state representation $|n\rangle$ of a system with $n$ particles. 
+
+One can define a basis of the Fock space with `FockBasis(N, offset=0)`, where `N` is the highest available Fock state and `offset` is the lowest cutoff state, which is by default zero. In the following example, we create a `FockState` with 3 quanta in an infinite-dimension Fock space:
```
Krastanov, line 11: The symbolic objects probably should always use `FockBasis(Inf)`. The non-inf is for cutoffs when doing numerical representations, where it has to be finite. That should be mentioned here and addressed below where we talk about expressing numerically.

docs/src/fock.md
```diff
@@ -0,0 +1,147 @@
...
+- Displacement operator: `DisplaceOp(alpha::Number, basis::Basis)`.
+
+## Numerical Conversions to QuantumOptics.jl
+
+Fock systems can be translated to the ket representation with `express`. For instance:
+
+```jldoctest
+julia> using QuantumOptics
+
+julia> b = FockBasis(3);
```
Krastanov, line 112: I think it makes more sense to not permit the creation of symbolic expressions with FockBasis with finite cutoff. These have only numeric meaning, symbolically a FockBasis has to be infinite. Rather the translation should have a configuration for how to do the cutoff.

ext/QuantumOpticsExt/QuantumOpticsExt.jl
```diff
@@ -70,28 +71,67 @@ express_nolookup(s::XBasisState, ::QuantumOpticsRepr) = (_s₊,_s₋)[s.idx]
...
+function express_nolookup(s::FockState, r::QuantumOpticsRepr)
+    b = basis(s)
+    i = s.idx
+    if !samebases(b, inf_fock_basis)
+        fockstate(b, i)
+    else
+        @warn "Fock space cutoff is not specified so we default to 2"
+        return fockstate(_bf2, i)
+    end
+end
```
Krastanov, line 83: Let's **not** use the cutoff in `basis(symbolic_fock_expression)`. Symbolic expressions should always have an infinite cutoff, otherwise we have to make difficult decisions about what bases are equal or not, how to perform simplifications, etc.

Rather, we should add a field to `QuantumOpticsRepr`, e.g. `fock_cutoff::Union{Nothing,2}` that is used here. The decided cutoff should not be a property of the symbolic expression, rather a property of how the conversion is happening.

src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -0,0 +1,133 @@
...
+const F₁ = const F1 = FockState(1,inf_fock_basis)
+
+##
+# Gates and Operators on harmonic oscillators
+##
+
+abstract type AbstractSingleBosonOp <: Symbolic{AbstractOperator} end
+abstract type AbstractSingleBosonGate <: AbstractSingleBosonOp end # TODO maybe an IsUnitaryTrait is a better choice
+isexpr(::AbstractSingleBosonGate) = false
+basis(x::AbstractSingleBosonOp) = x.basis
```
Krastanov, line 32: this should just hardcode `inf_fock_basis` like for the gate case below

Krastanov (changes requested):
I have a bunch of questions related to some of the details of how numerical expressions happen. I think I prefer we follow different semantics for the cutoff. Let's consider these notes as a first round of discussion on the topic.

docs/src/fock.md
apkille: Yes there is, good suggestion.

docs/src/fock.md
apkille: OK that sounds good to me.

docs/src/fock.md
apkille: Actually, I was mistaken. I thought there was something of this nature in SymbolicUtils.jl, but there isn't.

ext/QuantumOpticsExt/QuantumOpticsExt.jl
apkille: The problem with raising a warning in the constructor is that it creates allocations and increases the runtime. This causes tests in `test_express_opt.jl` to not pass. I instead added a warning to the docs page; that should be enough, right?

src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -0,0 +1,133 @@
...
+
+```jldoctest
+julia> c = CoherentState(im, FockBasis(4))
+|im⟩
+
+julia> phase = PhaseShiftOp(pi, FockBasis(4))
+U(π)
+
+julia> qsimplify(phase*c, rewriter=qsimplify_fock)
+|1.2246467991473532e-16 - 1.0im⟩
```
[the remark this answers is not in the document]
apkille: Same answer as the one for symbolic pi: it doesn't seem like the symbolics ecosystem has any support for this.

docs/src/fock.md
apkille: Maybe it would be good in the future to have boson and fermion bases in QuantumInterface?

ext/QuantumOpticsExt/QuantumOpticsExt.jl
```diff
@@ -1,13 +1,14 @@
...
 
 using QuantumInterface, QuantumOpticsBase
+using QuantumInterface: samebases
 using QuantumSymbolics
 using QuantumSymbolics:
     HGate, XGate, YGate, ZGate, CPHASEGate, CNOTGate, PauliP, PauliM,
     XCXGate, XCYGate, XCZGate, YCXGate, YCYGate, YCZGate, ZCXGate, ZCYGate, ZCZGate,
     XBasisState, YBasisState, ZBasisState,
     NumberOp, CreateOp, DestroyOp,
-    FockBasisState,
```
Krastanov, line 10: This is a breaking change, one that might even already affect QuantumSavory. Could you check if this is used in QuantumSavory? If it is not, then everything is fine (although it should this be accompanied by a breaking version bump and a note about breakage in the changelog)

src/QSymbolicsBase/QSymbolicsBase.jl
```diff
@@ -27,7 +27,7 @@ export SymQObj,QObj,
        I,X,Y,Z,σˣ,σʸ,σᶻ,Pm,Pp,σ₋,σ₊,
        H,CNOT,CPHASE,XCX,XCY,XCZ,YCX,YCY,YCZ,ZCX,ZCY,ZCZ,
        X1,X2,Y1,Y2,Z1,Z2,X₁,X₂,Y₁,Y₂,Z₁,Z₂,L0,L1,Lp,Lm,Lpi,Lmi,L₀,L₁,L₊,L₋,L₊ᵢ,L₋ᵢ,
-       vac,F₀,F0,F₁,F1,
+       vac,F₀,F0,F₁,F1,inf_fock_basis,
```
Krastanov, line 30: let's not export this for now (it makes it difficult to do non-breaking changes in the future)

src/QSymbolicsBase/express.jl
```diff
@@ -78,7 +78,17 @@ end
 ##
 
 """Representation using kets, bras, density matrices, and superoperators governed by `QuantumOptics.jl`."""
-struct QuantumOpticsRepr <: AbstractRepresentation end
+struct QuantumOpticsRepr <: AbstractRepresentation 
+    fock_cutoff::Union{Nothing, Int}
+    function QuantumOpticsRepr(fock_cutoff::Union{Nothing, Int})
```
Krastanov, line 83: This should probably not be an inner constructor. I also made a bunch of breaks to the documentation that currently would not work, but basically my goal is to make this a keyword argument constructor, so that it is easy to add more parameters to it in the future.

ext/QuantumOpticsExt/QuantumOpticsExt.jl
```diff
@@ -70,34 +71,14 @@ express_nolookup(s::XBasisState, ::QuantumOpticsRepr) = (_s₊,_s₋)[s.idx]
...
+express_nolookup(o::CreateOp, r::QuantumOpticsRepr) = create(FockBasis(r.fock_cutoff))
+express_nolookup(o::DestroyOp, r::QuantumOpticsRepr) = destroy(FockBasis(r.fock_cutoff))
+express_nolookup(o::DisplaceOp, r::QuantumOpticsRepr) = displace(FockBasis(r.fock_cutoff), o.alpha)
 express_nolookup(x::MixedState, ::QuantumOpticsRepr) = identityoperator(basis(x))/length(basis(x)) # TODO there is probably a more efficient way to represent it
-function express_nolookup(x::IdentityOp, ::QuantumOpticsRepr)
-    b = basis(x)
-    if b!=inf_fock_basis
-        return identityoperator(basis(x)) # TODO there is probably a more efficient way to represent it
-    else
-        @warn "Fock space cutoff is not specified so we default to 2"
```
Krastanov, line 97: there should still be a warning somewhere if the default cutoff is being used (maybe the warning can be moved to the constructor of `QuantumOpticsRepr`)

Krastanov (changes requested):
This is pretty good. I have the following requests:

- bring back the warning on default cutoff
- make the cutoff a keyword arg in the constructor (to make it easy to extend in the future)
- remove one of the exports
- check whether the naming change affects QuantumSavory and make sure the breaking change is mentioned in a changelong and an appropriate version bump is made

Feel free to merge without another review after this is done.

Krastanov:
let me know if I am missing something that is making the requests unreasonable

src/QSymbolicsBase/express.jl
apkille: Sure, thats OK with me.

apkille:
All of your requests sound good to me.

Krastanov:
ah, noted! Sounds good, let's keep it as a warning to the docs for now

Krastanov (changes requested):
I agree to the plan about adding a warning to the docs.

apkille:
@krastanov copying this down below: in your commit you modified `QuantumOpticsRepr` so that it has optional arguments, yet in your review you said that it should have a keyword argument. Could you please clarify?

PR #74, opened 2024-08-03
Solve piracies and ambiguities

Tortar:
Part of https://github.com/QuantumSavory/QuantumSavory.jl/pull/143

- [ ] The code is properly formatted and commented.

### DIFF
--- a/src/QSymbolicsBase/express.jl
+++ b/src/QSymbolicsBase/express.jl
@@ -64,6 +64,11 @@ express_nolookup(x, repr::AbstractRepresentation, ::UseAsState) = express_nolook
 # where the cache is a distribution over possible samples.
 express_from_cache(cache) = cache
 
+function consistent_representation(regs,idx,state)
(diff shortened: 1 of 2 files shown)


src/QSymbolicsBase/express.jl
```diff
@@ -64,6 +64,11 @@ express_nolookup(x, repr::AbstractRepresentation, ::UseAsState) = express_nolook
 # where the cache is a distribution over possible samples.
 express_from_cache(cache) = cache
 
+function consistent_representation(regs,idx,state)
```
Krastanov, line 67: `regs` is not something that QuantumSymbolics knows about. Any chance this can remain in QuantumSavory but maybe be annotated with `regs::VectorOrTupleOrWhatever{QuantumSavory.Register}`?

In particular `r.reprs` at line 68 is an implementation detail that comes from `QuantumSavory`

Tortar: seems better, for now then this PR is no longer necessary

PR #76, opened 2024-08-11
Move docstrings to struct definition so that we do not export undocumented symbols

Krastanov:
@apkille , just FYI, undoing a small change of yours that created some downstream documentation/testing issues. 

Once 1.11 is out we should just add this to the tests https://github.com/JuliaLang/julia/pull/52413


Krastanov:
> Once 1.11 is out we should just add this to the tests [JuliaLang/julia#52413](https://github.com/JuliaLang/julia/pull/52413)

That is not necessary... It seems this is enough:

```
    checkdocs = :exports,
    warnonly = false,
```

apkille:
@Krastanov this looks good to me, thanks for doing this!

PR #77, opened 2024-08-11
Modify @withmetadata to filter strings in field definitions

apkille:
Addresses errors in https://github.com/QuantumSavory/QuantumSavory.jl/pull/144.


Krastanov:
ah, interesting! These are for inline documentation of struct fields. Thanks for fixing this

PR #79, opened 2024-08-13
Updated operators to print with hat

rbaliarizona:
Operators now print with hat on top


rbaliarizona:
@Krastanov  my main goal here was to familiarize myself with the pull-request process. I personally like my operators to display with hats, but I know this isn't a law, so feel free to discard this change. Any feedback on if I went through the pull request process would be welcome though :)

Krastanov:
This is great, Rohan, thanks! I will run a review on it later today

src/QSymbolicsBase/literal_objects.jl
```diff
@@ -138,7 +138,7 @@ basis(x::SymQ) = x.basis
 
 Base.show(io::IO, x::SKet) = print(io, "|$(symbollabel(x))⟩")
 Base.show(io::IO, x::SBra) = print(io, "⟨$(symbollabel(x))|")
-Base.show(io::IO, x::Union{SOperator,SHermitianOperator,SUnitaryOperator,SHermitianUnitaryOperator,SSuperOperator}) = print(io, "$(symbollabel(x))")
+Base.show(io::IO, x::Union{SOperator,SHermitianOperator,SUnitaryOperator,SHermitianUnitaryOperator,SSuperOperator}) = print(io, "$(symbollabel(x))̂")
```
Krastanov, line 141: Two points:

- we probably should insert the hat on the first letter (not all operators are a single symbol)

- let's remove `SSuperOperator` from this list for the moment, as it has a bit more complicated printing

Krastanov (changes requested):
I love that the PR is small -- way easier to review and merge.

I left one inline comment on some code details.

There are a lot of failing tests. An easy way to run the tests locally and check things is to do `] test QuantumSymbolics` -- that way you can find exactly which test is failing and focus on it.

Most of the tests here seem to be "doctests" that just lack a hat. There is an automated way to fix them (the `fix` argument to `doctest` function).

Happy to discuss testing more at the hackathon.

rbaliarizona:
@Krastanov thanks for the feedback!

I have updated the doctests. Hopefully they pass now. 

I actually added logic to make the hat print on the middle character instead. I know this wasn't your suggestion and it may be more logic that we are looking for here. Lmk what you think!

Krastanov:
apologies, it seems I messed up the merge with master. Let me try to fix that

Krastanov:
Darn, there are a few new places in the documentation where the hats need to be added. For now I will mark this as a draft as I will not have the time to make these changes. Apologies for the delay. Happy to pick it back up if you are interested in finishing it, but no pressure on that.

rbaliarizona:
Hi @Krastanov , thanks for looking at it. Unfortunately I think I will need to drop this, atleast for now.

Krastanov:
no worries!

PR #85, opened 2024-10-19
fix Aqua's reported piracies and method ambiguities

Omar-Elrefaei:
Fix all of Aqua's reported piracies and method ambiguities. (#65)

Precisely 37 out of the 39 potential piracies are avoided by informing Aqua that types from QuantumInterface are in fact our own. The last two are fixed by adjusting the `* and +` functions that accept a `Vararg`, it had a problem for the case when `N=0`. This solution avoids both the unbound_type_param and the piracy, and seems to be a reasonable design to not allow zero arguments passed to the function (similar to Base operators themselves)

<details>
<summary>PR checklist</summary>

- [ ] The code is properly formatted and commented.
- [ ] Substantial new functionality is documented within the docs.
- [] All new functionality is tested.
- [ ] All of the automated tests on github pass.
- [ ] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them. <small>There will be plenty of old code that is flagged as we are slowly transitioning to enforced formatting. Please do not worry about or address older formatting issues -- keep your PR just focused on your planned contribution.</small>
</details>

### DIFF
--- a/test/test_aqua.jl
+++ b/test/test_aqua.jl
@@ -1,7 +1,9 @@
 @testitem "Aqua" tags=[:aqua] begin
     using Aqua
+    import QuantumInterface as QI
+     own_types = [QI.AbstractBra, QI.AbstractKet, QI.AbstractSuperOperator, QI.AbstractOperator]
     Aqua.test_all(QuantumSymbolics,
-            ambiguities=(;broken=true),
-            piracies=(;broken=true),
+            ambiguities=(),
+            piracies=(;treat_as_own=own_types),
     )
 end
--- a/test/test_aqua.jl
+++ b/test/test_aqua.jl
@@ -1,9 +1,42 @@
 @testitem "Aqua" tags=[:aqua] begin
     using Aqua
-    import QuantumInterface as QI
-     own_types = [QI.AbstractBra, QI.AbstractKet, QI.AbstractSuperOperator, QI.AbstractOperator]
-    Aqua.test_all(QuantumSymbolics,
-            ambiguities=(),
-            piracies=(;treat_as_own=own_types),
-    )
+    
+    # Add any new types needed to QObj, or here if QObj if not appropriate.
+    # Add types from elsewhere in the ecosystem here or preferably to QObj
+    own_types = [Base.uniontypes(QObj)...,]
+    own_types_union = Union{SymQObj,}
+
+    Aqua.test_all(QuantumSymbolics, piracies=(;treat_as_own=own_types))
+
+    function normalize_arguments(method)
+        args = Base.unwrap_unionall(method.sig).types[2:end]
+        normalized_args = []
+        # handle few edge cases specific to our analysis
+        for arg in args
+            # mutation and order of if-conditions is intedtional here
+            if (arg isa UnionAll) && (arg.body <: Type) arg = arg.body.parameters[1] end
+            if (arg isa Core.TypeofVararg) arg = arg.T end
+            if (arg isa TypeVar) arg = arg.ub end
+            push!(normalized_args, arg)
+        end
+        return normalized_args
+    end
+
+    # Custom type-piracy detection, to catch uses of QuantumInterface types without a Symbolic
+    filtered_piracies = filter(Aqua.Piracy.hunt(QuantumSymbolics)) do m
+        !any(normalize_arguments(m) .<: own_types_union)
+    end
+
+    aqua_piracies = Aqua.Piracy.hunt(QuantumSymbolics, treat_as_own=own_types)
+    internally_detected_piracies = setdiff(filtered_piracies, aqua_piracies)
+    if !isempty(internally_detected_piracies)
+        printstyled(
+            stderr,
+            "Internally flagged possible type-piracy:\n";
+            color = Base.warn_color()
+        )
+        show(stderr, MIME"text/plain"(), internally_detected_piracies)
+        println(stderr, "\n")
+    end
+    @test isempty(internally_detected_piracies)
 end
--- a/src/QSymbolicsBase/basic_ops_homogeneous.jl
+++ b/src/QSymbolicsBase/basic_ops_homogeneous.jl
@@ -29,7 +29,7 @@ arguments(x::SScaled) = [x.coeff,x.obj]
 operation(x::SScaled) = *
 head(x::SScaled) = :*
 children(x::SScaled) = [:*,x.coeff,x.obj]
-function Base.:(*)(c, x::Symbolic{T}) where {T<:QObj}
+function Base.:(*)(c::U, x::Symbolic{T}) where {U<:Union{Number, Symbolic{<:Number}},T<:QObj}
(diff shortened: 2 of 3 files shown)


test/test_aqua.jl
```diff
@@ -1,7 +1,9 @@
 @testitem "Aqua" tags=[:aqua] begin
     using Aqua
+    import QuantumInterface as QI
+     own_types = [QI.AbstractBra, QI.AbstractKet, QI.AbstractSuperOperator, QI.AbstractOperator]
```
Krastanov, line 4: Could you elaborate why these need to be marked as `own`? I am not sure whether this would be consistent with how independent packages downstream of QuantumInterface (e.g. QuantumOptics or QuantumCummulants) -- I want to make sure there are some precautions put in place for potential future issues.

Krastanov (commented):
Thank you for taking care of this! I have some minor questions about the types marked as `own` but on a quick skim the rest looks good.

Krastanov:
Hi! Just a friendly bump on this. I see most of the piracies were masked by marking a bunch of types as owned. This will be a bit fragile. Do you have any insight into which piracies this addresses -- upstreaming them to QuantumInterface would be a safer long-term solution.

Omar-Elrefaei:
Hi @Krastanov, I replied to your code review itself a while ago. Maybe GitHub doesn't show it too clearly https://github.com/QuantumSavory/QuantumSymbolics.jl/pull/85#discussion_r1827075621

Krastanov:
No worries, I have made the same mistake in the past.

Krastanov:
I see, that makes sense, thank you!

So an example of an offender would be

```
Base.:(*)(x::Symbolic{T}, y::Symbolic{S}) where {T<:QObj,S<:QObj}
```

I am wondering, can `own_types` take something like `Symbolic{<:QI.AbstractBra}`, in order to be more specific about what types are actually owned? After all, given the premise of this library, we can not really claim ownership of `QI.AbstractBra`, but `Symbolic{<:QI.AbstractBra}` is very much "owned".

Omar-Elrefaei:
I'm won't be exactly sure why, but that doesn't seem to work.

At the simplest case `Aqua.test_piracies(QuantumSymbolics, treat_as_own = [Symbolic{QI.AbstractOperator})` still shows the following. (also tried with `<:QI.AbstractOperator`)
```
Possible type-piracy detected:
......
[35] tr(x::Symbolic{QuantumInterface.AbstractOperator}) @ QuantumSymbolics ~/.julia/dev/QuantumSymbolics/src/QSymbolicsBase/linalg.jl:305
[37] vec(x::Symbolic{QuantumInterface.AbstractOperator}) @ QuantumSymbolics ~/.julia/dev/QuantumSymbolics/src/QSymbolicsBase/linalg.jl:527
```

Omar-Elrefaei:
> After all, given the premise of this library, we can not really claim ownership of `QI.AbstractBra`

The way I'm thinking about it is that QuantumInterface is that one that owns `AbstractBra` in reality. If a user comes along and instantiates a `Vector{<:AbstractBra}` object, they still don't own it, and defining `adjoint` on it would still be a piracy. 

What we are saying here is that QuantumSymbolics gets a special treatment because it is really an extension of QuantumInterface by the same developers. QuantumInterface itself says `This package is not meant for public use!` in the readme.  I'm assuming non of your other quantum packages will ever be interested in defining `Base.:(*)` too.

Krastanov:
You are right, it is just that there are multiple groups coordinating on this work, and having as much of the coordination automated would help a lot.

In particular, if we just `treat_as_own` the bare `AbstractKet`, we will miss actual issues like `f(::AbstractKet)` which should be in `QuantumInterface` not in `QuantumSymbolics`, making Aqua less helpful than it could be -- this is my main worry.

I posted about this on slack. Let's wait a bit to see if anyone has suggestions there.

Krastanov:
Actually, how about we keep your implementation as is, but also add the following extra test?

```
all([m.sig.types[2:end] ⊆ own_types for m in Aqua.Piracy.hunt(QuantumSymbolics)])
```

The subset check might be too harsh. An intersection check might be enough.

Omar-Elrefaei:
> we will miss actual issues like f(::AbstractKet)

Ok, that is the use case I didn't think of on my own

> but also add the following extra test?

Yeah, doesn't seem like a bad idea. I went through few iterations to get to something that doesn't show false-positives.

```julia
# get it to work without errors
all([Base.unwrap_unionall(m.sig).types[2:end] ⊆ own_types for m in Aqua.Piracy.hunt(QuantumSymbolics)])

get_method_args(m::Method) = Base.unwrap_unionall(m.sig).types[2:end] 

# Transform to filter to see the cases that are not passing
filter(m -> !(get_method_args(m) ⊆ own_types), Aqua.Piracy.hunt(QuantumSymbolics)) # 18 false-positives

# Use intersection test. Subset is too strict because it keeps out methods which take additional arguments
#not related to QuantumSymbolics.  Example ptrace(x::Symbolic{QuantumInterface.AbstractOperator}, s)
filter(m -> isdisjoint(get_method_args(m), own_types), Aqua.Piracy.hunt(QuantumSymbolics)) #10 false-positives

# Some arguments need a little normalization. Such as maketerm(::Type{<:Symbolic{<:QObj}},...)
function normalize(arg)
    if (arg isa UnionAll) && (arg.body <: Type) arg = arg.body.parameters[1] end
    if (arg isa Core.TypeofVararg) arg = arg.T end
    if (arg isa TypeVar) arg = arg.ub end
    return arg
end

filter(m -> isdisjoint(normalize.(get_method_args(m)), own_types), Aqua.Piracy.hunt(QuantumSymbolics)) #8 false-positives

# What we really need is to check if an arg is a subtype of Symbolic{<:QObj}, to be able to match most edge cases
filter(m -> !any(normalize.(get_method_args(m)) .<: Symbolic{<:QObj}), Aqua.Piracy.hunt(QuantumSymbolics)) # 0 false-positives
```

Here is a proof of it catching an introduced piracy
```julia
module modA
       import QuantumInterface: AbstractBra, AbstractKet
       import Base.println
       import SymbolicUtils.Symbolic
       Base.println(x::T) where T<:Union{AbstractBra, AbstractKet} = 1
       Base.println(x::Symbolic{T}) where T<:Union{AbstractBra, AbstractKet} = 2
end

filter(m -> !any(normalize.(get_method_args(m)) .<: Union{Symbolic{<:QObj}}), Aqua.Piracy.hunt(modA))
#[1] println(x::T) where T<:Union{AbstractBra, AbstractKet} @ Main.modA REPL[53]:5
```

Krastanov:
This is wonderfully done, thank you for working it out!

Omar-Elrefaei:
I'm good with this. Might have gotten carried away.

Krastanov:
Thank you, this is very much appreciated.

Could you message me at skrastanov@umass.edu with your name, github username, and bounty number so I can forward you the bounty payment details?

PR #86, opened 2024-11-12
Bump compat for Symbolics to 6 and fix maketerm arguments

akirakyle:
Supersedes #78 fixing the failed tests introduced by a change in the interface for `maketerm` in TermInterface.jl


Krastanov:
The downgrade check seems to be failing. Maybe because now the compat should be only for 6? I will do that quick check.

Krastanov:
Thanks! Will be released shortly

PR #87, opened 2024-11-12
Latexify AbstractSingleBosonOp instead of AbstractSingleBosonGate

akirakyle:
Fixes latexify of expressions containing subtypes of `AbstractSingleBosonOp` such as `NumberOp`, `CreateOp`, `DestroyOp`.

Relatedly it doesn't look like `AbstractSingleBosonGate` is used anywhere, and the comment seems like it may be outdated since it seems like implementing `isunitary` is the way such checks are handled?


Krastanov:
yup, we do have some outdated types that we should probably just drop

Thanks for the fix!

Issue #92, opened 2024-12-22
Move `express` and representation types to QuantumInterface


apkille:
Right now, I'm working on translating phase space representations of Gaussian quantum objects in Gabs.jl to Fock space representations in QuantumOptics.jl. Should we limit `express` to one way symbolic-to-numerical translations, or could we use it as a general function that translates between any two packages that depend on QuantumInterface.jl? To me, it seems pretty clean to implement something like `express(::GaussianState, QuantumOpticsRepr(cutoff = 10))` and return a QuantumOptics `Ket` object with `FockBasis(10)`.

Krastanov:
I am in favor of this.

QuantumClifford already has an extension that converts some objects to QuantumOptics objects, but those are unambiguous and do not need configuration like `cutoff=10`. For the more general objects you are describing, this seems like the appropriate way to proceed.

PR #93, opened 2024-12-22
Migrate `express` functionality and `AbstractRepresentation` type to QuantumInterface

apkille:
#92 https://github.com/qojulia/QuantumInterface.jl/pull/46
The problem with this PR and the one in QuantumInterface is that a type piracy is introduced here:
```
express(s, repr::AbstractRepresentation) = express(s, repr, UseAsState())
```
I suppose we could migrate `AbstractUse` and its subtypes to QuantumInterface as well, but 
```
function express(state::Symbolic, repr::AbstractRepresentation, use::AbstractUse)
```
would then be a type piracy. Plus `AbstractUse` is intended for dealing with edge cases in symbolic-to-Clifford translations, so to me it doesn't make sense to migrate it up to QuantumInterface.


apkille:
@Krastanov

src/QSymbolicsBase/QSymbolicsBase.jl
```diff
@@ -53,8 +54,6 @@ export SymQObj,QObj,
 # Metadata cache helpers
 ##
 
-"""An abstract type for the supported representation of quantum objects."""
-abstract type AbstractRepresentation end
 abstract type AbstractUse end
 struct UseAsState <: AbstractUse end
 struct UseAsOperation <: AbstractUse end
```
Krastanov, line 59: Let's move these too, unless there is a reason not to.

src/QSymbolicsBase/express.jl
```diff
@@ -37,8 +37,6 @@ julia> express(QuantumSymbolics.X, CliffordRepr(), UseAsObservable())
 + X
 ```
 """
-function express end
```
Krastanov, line 40: Could you update the docstring above, so it is clear this is about symbolics, not about all of the express capabilities? Just adding the work "symbolic" somewhere would be enough.

Krastanov (changes requested):
Looks good! Three minor things:

- The comment about `AbstractUse` -- I think it makes sense to move this too. E.g. if QuantumClifford is converting an Operator it needs to know whether the user wants a Stabilizer or a CliffordOperator
- The docstring update (or move)
- Check that `express` still shows in the documentation -- now that it does not belong to QuantumSymbolics, it might need to be explicitly included.

Otherwise feel free to merge once CI passes, the Project.toml is bumped, without waiting for another review. I guess that requires the other PR in QuantumInterface to be merged first.

Krastanov:
Marking it as a draft to remove it from my review queue for now. Convert it back when ready and merge (or ping me to merge).

apkille:
> * Check that `express` still shows in the documentation -- now that it does not belong to QuantumSymbolics, it might need to be explicitly included.
 
It still shows in the docs.

PR #95, opened 2025-01-01
Rotation Gates

### LINKED ISSUE #94: Rotation Gates
Rotation Gates are not available in QuantumSymbolics.

I have implemented Rotation Gates in a fork: https://github.com/sagnikpal2004/QuantumSymbolics.jl/tree/rgate
I have tested the feature, and it works well. The unit tests also pass.

Here is its usage:
```
using QuantumSymbolics

express(Rx(1.0*pi))
express(Ry(1.0*pi))
express(Rz(1.0*pi))

express(RGate(:x, 0.5*pi))
QuantumSavory.apply!(reg[1], Ry(0.25*pi))
```

Let me know if this is something that can be merged into QuantumSymbolics

sagnikpal2004:
QuantumSymbolics.jl is currently missing rotation gates.
I decided to implement these gates as RGate, which are structs which take in a direction and angle.
`express` functionality for QuantumOpticsRepr is also implemented 

CHANGELOG:
- Added Rotation Gates: RGate, Rx, Ry, Rz
- Implemented `express` in QuantumOpticsRepr

CHECKLIST:
- [X] The code is properly formatted and commented.
- [ ] Substantial new functionality is documented within the docs.  <b>--TODO</b>
- [X] All new functionality is tested.
- [X] All of the automated tests on github pass.
- [X] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them. <small>There will be plenty of old code that is flagged as we are slowly transitioning to enforced formatting. Please do not worry about or address older formatting issues -- keep your PR just focused on your planned contribution.</small>

Edit from Stefan: closes #94


src/QSymbolicsBase/predefined.jl
```diff
@@ -106,6 +106,14 @@ symbollabel(::HGate) = "H"
 ishermitian(::HGate) = true
 isunitary(::HGate) = true
 
+@withmetadata struct RGate <: AbstractSingleQubitGate 
+    dir::Symbol
+    θ::Float64
+end
```
Krastanov, line 112: A few points:

- this object does not seem to be a "leaf" of a symbolic expression, and as such it needs to implement additional APIs that permit symbolic tree traversal
- theta would not necessarily be a float64. It might also be a symbolic scalar, a float32, etc. There are a lot of sophisticated approaches to what this can be set to, but for the moment it would probably work best if it is unconstrained
- the dir symbol should probably just be removed (and then have three different types RotX RotY RotZ) -- otherwise it will be very difficult to write simplification rules for these objects

See the "scaled" object as an example of the additional API that is needed for non-leaf symbolic objects https://github.com/QuantumSavory/QuantumSymbolics.jl/blob/main/src/QSymbolicsBase/basic_ops_homogeneous.jl#L22

Krastanov: if you want to, you can always create a `AbstractRotGate` if you want to avoid some minor code repetition

Krastanov (changes requested):
This is a great start, thank you for engaging with this.

I am leaving a few comments on suggested structure below.

We should probably also add some simplification rules to the database of rules: https://github.com/QuantumSavory/QuantumSymbolics.jl/blob/main/src/QSymbolicsBase/rules.jl

It can be a new `_ROT` list of rules, containing things like `Rot(a)*Rot(b) = Rot(a+b)` and commutation relationships between them and commutation relationships with Paulis, and conversion of exponentials of Paulis into rot gates.

For testing we can add something that checks an equality of a bunch of expressions of the type `express(expr) == express(expr(simplify(expr)))` -- that way we check consistency between the numerical representation and the symbolic simplification rules.

sagnikpal2004:
Summary of changes made
- Added abstract type AbstractRotGate
- Created three different gates RotXGate, RotYGate, RotZGate (and aliases Rx, Ry, Rz)
- Implemented TermInterface APIs for the gates
- Added rules for rotation gate simplification
- Added tests for rule application

src/QSymbolicsBase/rules.jl
```diff
@@ -109,7 +109,22 @@ RULES_FOCK = [
     @rule(~o::_isa(SqueezeOp) * ~k::_isequal(vac) => SqueezedState((~o).z, (~o).basis))
 ]
 
-RULES_SIMPLIFY = [RULES_PAULI; RULES_COMMUTATOR; RULES_ANTICOMMUTATOR; RULES_FOCK]
+RULES_ROT = [
+    @rule(~r::_isa(RotX) => I where (~r).θ == 0),
+    @rule(~r::_isa(RotY) => I where (~r).θ == 0),
+    @rule(~r::_isa(RotZ) => I where (~r).θ == 0),
```
Krastanov: these should probably be handled by the constructor

src/QSymbolicsBase/rules.jl
```diff
@@ -109,7 +109,22 @@ RULES_FOCK = [
...
+    @rule(~r1::_isa(RotX) * ~r2::_isa(RotX) => try RotX((~r1).θ + (~r2).θ) catch end),
+    @rule(~r1::_isa(RotY) * ~r2::_isa(RotY) => try RotY((~r1).θ + (~r2).θ) catch end),
+    @rule(~r1::_isa(RotZ) * ~r2::_isa(RotZ) => try RotZ((~r1).θ + (~r2).θ) catch end),
+    @rule(~r::_isa(RotX) => try RotX(mod((~r).θ, 2π)) catch end),
+    @rule(~r::_isa(RotY) => try RotY(mod((~r).θ, 2π)) catch end),
+    @rule(~r::_isa(RotZ) => try RotZ(mod((~r).θ, 2π)) catch end),
+    @rule(exp(~α * ~x::_isa(XGate)) => try real(~α) == 0 ? RotX(-2imag(~α)) : nothing catch end),
+    @rule(exp(~α * ~x::_isa(YGate)) => try real(~α) == 0 ? RotY(-2imag(~α)) : nothing catch end),
+    @rule(exp(~α * ~x::_isa(ZGate)) => try real(~α) == 0 ? RotZ(-2imag(~α)) : nothing catch end)
```
Krastanov: What does it mean to have a `try catch` that returns nothing on an error in a simplification rule? Could you give an example of how this works?

src/QSymbolicsBase/predefined.jl
```diff
@@ -143,6 +174,12 @@ const Pm = const σ₋ = PauliM()
 const Pp = const σ₊ = PauliP()
 """Hadamard gate"""
 const H = HGate()
+"""Rotation around the X axis"""
+const Rx = RotXGate
+"""Rotation around the Y axis"""
+const Ry = RotYGate
+"""Rotation around the Z axis"""
+const Rz = RotZGate
```
Krastanov, line 182: These aliases seem a bit out of place here. All the other consts here are not aliases, rather instances of types. Aliases could be useful, but for now let's just stick to only the types, to avoid accidental complexity.

src/QSymbolicsBase/predefined.jl
```diff
@@ -106,6 +106,37 @@ symbollabel(::HGate) = "H"
...
+isexpr(::AbstractRotGate) = true
+iscall(::AbstractRotGate) = true
+arguments(x::AbstractRotGate) = [x.θ]
+
+@withmetadata struct RotXGate <: AbstractRotGate
+    θ
+end
+operation(::RotXGate) = Rx
+head(::RotXGate) = :RotX
+children(x::RotXGate) = [:RotX, x.θ]
```
Krastanov, line 121: This `head` (`:RotX`) does not seem to be something that exists. This (and for Y and for Z) should probably be `:RotXGate`

Krastanov (changes requested):
Thanks for updating this! There are a couple of simplification rules that I do not fully understand, could you address the questions below?

Are there predefined rotations gates in QuantumOptics.jl? If yes, we should probably use them or at least make sure we are consistent with them.

Krastanov:
Marking as draft just so I can keep my review queue organized. Do not hesitate to mark it back as ready and re-request a review.

src/QSymbolicsBase/rules.jl
sagnikpal2004: From the documentation of [SymbolicUtils.Chain](https://symbolicutils.juliasymbolics.org/api/#:~:text=Chain(itr)), if the return value of a rule rewriter is `nothing`, then the rule is not applied, and the original expression remains unchanged.
I used `try...catch` here, just in case, functions like `mod` or `real` or `imag` are not defined for some symbolics. It works perfectly when I test it with `<:Number`, and with anything else, it keeps the expression unchanged.

src/QSymbolicsBase/predefined.jl
```diff
@@ -106,6 +106,43 @@ symbollabel(::HGate) = "H"
...
+
+@withmetadata struct RotZGate <: AbstractRotGate
+    θ
+end
+operation(::RotZGate) = Rz
+head(::RotZGate) = :Rz
+children(x::RotZGate) = [:Rz, x.θ]
+symbollabel(x::RotZGate) = "Rz($(x.θ))"
+"""Rotation around the Z axis"""
+Rz(θ) = (θ == 0) ? I : RotZGate(θ)
```
Krastanov, line 144: I think we do not need separate functions for this type.

- note that in the CNOT case, it is an **instance**, not a new renamed constructor
- note that in the symbolic trace case, it is a way to hook into an existing LinearAlgebra API, so that is why we have a renamed constructor -- no-preexisting API to hook into here.

sagnikpal2004: Removed these functions.
However, the `@withmetadata` macro does not work with constructors inside the struct
The `$\theta$ == 0` rule has been added to `rules.jl`

Issue #64, opened 2024-07-09
Thorough benchmarks [$200]


Krastanov:
<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

Currently the project is claimed by `no one` until `...`.

If you want to, you can work on this project without making a claim, however claims are encouraged to give you and other contributors peace of mind. Whoever has made a claim takes precedence when solutions are considered.

You can always propose your own funded project, if you would like to contribute something of value that is not yet covered by an official bounty.
</details>

# Project: "Thorough benchmarks" [$200]

We have a small benchmark suite already implemented for our other repositories, which is executed as part of our CI runs. It is defined in the `benchmark` folder and reported for each pull request. **We would like to implement a similar benchmark suite here** to include many facets of this library. E.g. basic symbolic operations, simplifications, rule application, flattening and expanding, expressing into numerical backends, time to import, etc. The new benchmarks should be legible, easy to follow, and organized by topic. Most of them should be microbenchmarks testing only one concept, but a few holistic benchmarks would make sense as well.

**Required skills**: Generic Julia skills.

**Reviewer**: Stefan Krastanov

**Duration**: 1 month

#### Payout procedure:

The Funding for these bounties comes from the National Science Foundation and from the NSF Center for Quantum Networks. The payouts are managed by the NumFOCUS foundation and processed in bulk once every two months. If you live in a country in which NumFOCUS can make payments, you can participate in this bounty program.

[Click here for more details about the bug bounty program.](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

coderight1:
Hi @Krastanov , I would like to work on this .

Krastanov:
Hi, @coderight1 . Apologies for the slow response, I did not see this message originally. Do you have past experience experience with Julia? Could you list out a rough plan for what you are going to do?

Omar-Elrefaei:
I'm ready to take this on @Krastanov

Krastanov:
sounds great! Marking as reserved

PR #97, opened 2025-02-02
add package benchmarks

Omar-Elrefaei:
This is a draft for benchmarks as discussed here #64.

So far I covered
1. time to import, latency
2. simple object instantiations
3. arithmetic operations
4. linear algebra
5. symbolic manipulation (expand, simplify)
6. expressing in optics/clifford backend

Please check if I missed any major features.

If there is any real usage notebook you'd like me to extract usages from I'd be happy to. 
Also if you can hint me at specific directions of what a "hot loop" would look like, I could stress test that use case. I'm just not sure, for example, whether multiplication of hundreds of `kets` or `ops` is relevant or too contrived.

### DIFF
--- a/benchmark/benchmarks.jl
+++ b/benchmark/benchmarks.jl
@@ -18,13 +18,28 @@ ttfx_command = `
     """`
 SUITE["latency"]["using"] = @benchmarkable run(load_command) samples=3 seconds=15
 SUITE["latency"]["ttf_operation"] = @benchmarkable run(ttfx_command) samples=3 seconds=15
-SUITE["latency"]["ttf_simplify"] = @benchmarkable qsimplify(X*Y) samples=1
+SUITE["latency"]["ttf_simplify"] = @benchmarkable qsimplify(X*Y) samples=1 evals=1
 
-# Basic symbolic object creation
+
+# Symbolic object creation
 SUITE["creation"] = BenchmarkGroup(["symbolic"])
 SUITE["creation"]["ket"] = @benchmarkable @ket _k
 SUITE["creation"]["op"] = @benchmarkable @op _A
 SUITE["creation"]["super_op"] = @benchmarkable @superop _S
+
+SUITE["creation"]["large_trees"] = BenchmarkGroup(["allocs"])
+function large_tree_with_plenty_reallocations(layers)
+    expr_op = QuantumSymbolics.I
+    expr_ket = X1
+    for _ in 1:layers
+        expr_op = rand([X, Y, Z, H]) + (rand([X, Y, Z, H]) * expr_op)
+        expr_ket = (expr_op * expr_ket) + rand([X1, X2, Y1, Y2, Z1, Z2])
+    end
+    return expr_op, expr_ket
+end
+SUITE["creation"]["large_trees"]["10_layers"] = @benchmarkable large_tree_with_plenty_reallocations(10)
+SUITE["creation"]["large_trees"]["50_layers"] = @benchmarkable large_tree_with_plenty_reallocations(50)
+
 
 # Basic operations
 SUITE["operations"] = BenchmarkGroup(["symbolic"])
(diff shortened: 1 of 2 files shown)


benchmark/benchmarks.jl
```diff
@@ -0,0 +1,93 @@
...
+pauli_state_8 = (conj(XCY⊗Y) + dagger(Z⊗YCX) + transpose(0.5im*Y ⊗ exp(XCY))) * (X1⊗Y1⊗Z1)
+
+SUITE["express"]["optics"]["simple_op"] = @benchmarkable express($simple_op) setup=clear_cache!(simple_op) evals=1
+SUITE["express"]["optics"]["simple_ket"] = @benchmarkable express($simple_ket) setup=clear_cache!(simple_ket) evals=1
+SUITE["express"]["optics"]["pauli_op_4"] = @benchmarkable express($pauli_op_4) setup=clear_cache!(pauli_op_4) evals=1
+SUITE["express"]["optics"]["pauli_state_8"] = @benchmarkable express($pauli_state_8) setup=clear_cache!(pauli_state_8) evals=1
+
+# TODO: not sure what else to add here
+SUITE["express"]["clifford"]["simple_ket"] = @benchmarkable express($simple_ket, CliffordRepr()) setup=clear_cache!(simple_ket) evals=1
+SUITE["express"]["clifford"]["simple_observable"] = @benchmarkable express($simple_op, CliffordRepr(), UseAsObservable()) setup=clear_cache!(simple_op) evals=1
```
Omar-Elrefaei, line 92: What I had the most trouble with is expressing in Clifford formalism as it seems that most operations are not supported except for the most basic tensor products.

Krastanov (changes requested):
This looks great! Thank you for spending time on it!

The only other thing that would be valuable to include is something that tests issues related to the fact that we are not using abstract data types (like Unityper, SumTypes, Expronicon, or Moshi). Because of the lack of this feature, manipulating large trees is actually somewhat expensive and it would be valuable to keep track of that expense.

So a holistic benchmark for that can look something like:

```
function make_large_tree_with_plenty_of_reallocations(layers)
    expr_op = identity
    expr_ket = X1
    for _ in 1:layers
        expr_op = rand([X,Y,Z,H,...])+rand([...])*expr_op
        expr_ket = expr_op*expr_ket + rand([X1,X2,Z1,Z2,...])
    end
    return expr_op, expr_ket
end


SUITE["large_tree"]["construct"]["10_layers"] = make_large_tree_with_plenty_of_reallocations(10)
```

We are not including tensor products here or superoperators, but that is probably fine until we invest time in actually using abstract data types.

benchmark/benchmarks.jl
Krastanov: I think we can just stick to what you already have here. Expanding on the Clifford capabilities would need to be a separate task.

benchmark/benchmarks.jl
```diff
@@ -0,0 +1,93 @@
...
+load_command = `julia --quiet --project=./ --eval="using QuantumSymbolics"`
+ttfx_command = `
+    julia --quiet --project=./ --eval="""
+        using QuantumSymbolics;
+        @ket k1; @op A;
+        A * commutator(A,X) * k1
+    """`
+SUITE["latency"]["using"] = @benchmarkable run(load_command) samples=3 seconds=15
+SUITE["latency"]["ttf_operation"] = @benchmarkable run(ttfx_command) samples=3 seconds=15
+SUITE["latency"]["ttf_simplify"] = @benchmarkable qsimplify(X*Y) samples=1
```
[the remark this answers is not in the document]
Omar-Elrefaei: Oh, the qsimplify one definitely needs `evals=1`. because here we are measuring first-run inference/compile time and really need the assertion that we're only running it once. (assuming it is a new session, which makes me think this is not a very robust setup). 

`evals` is adjusted by `tune!` in cases where the runtime of one function evaluation is too short relative to the timing resolution. so not necessary for the first ones, but might as well for clarity

Krastanov:
thanks for the bump! This looks good. I will close and reopen to rerun the tests because there is something weird happening with them (maybe a break in a dependency?)

Krastanov:
sigh... it seems the error indeed exists, it did not use to exist a few weeks ago, it is unrelated to your changes, and it is probably related to something in dependencies.

Let me mark it as broken

Krastanov (approved):
Thank you for contributing this! The test failure is frustrating but it seems it is related to dependencies that exist only on julia 1.11. I will merge this later today after the rest of the tests are done. Please feel free to proceed with submitting your bounty invoice on open collective

PR #101, opened 2025-02-19
Fix launching of Benchmarks CI

Omar-Elrefaei:
Sorry @Krastanov I didn't notice last PR that the benchmarks wasn't really running.


Krastanov:
I did not notice this either. Thank you for following up on it!

PR #106, opened 2025-02-28
Remove excessive parenthesis in addition and tensor products

apkille:
```
@op A; @op B; @op C SpinBasis(1//2) ⊗ SpinBasis(1//2);
```
Before:
```
julia> A + B
(A+B)

julia> C + (A ⊗ B)
((A⊗B)+C)
```
After:
```
julia> A + B
A+B

julia> C + (A ⊗ B)
(A⊗B)+C
```
Same goes for bras, kets, and superoperators.

### DIFF
--- a/docs/src/introduction.md
+++ b/docs/src/introduction.md
@@ -78,7 +78,7 @@ Similar scaling procedures can be performed on bras and operators. Addition betw
 julia> @op A₁; @op A₂;
 
 julia> A₁+A₂
-(A₁+A₂)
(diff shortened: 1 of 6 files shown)


Krastanov (approved):
Thanks! Looks good to merge once test pass -- feel free to squash and merge without further review.

Krastanov:
if QuantumSavory checks fail, it would not be due to this PR. One of the buildkite runs is QuantumSavory, so do not block on that

Issue #108, opened 2025-04-21
Arrays of operators


jeffwack:
I working on simulating quantum systems described by [the SLH framework](https://arxiv.org/abs/1611.00375). I am interested in using QuantumSymbolics because I have the [same interest described here](https://github.com/qojulia/QuantumOpticsBase.jl/issues/70#issue-1419860820) in using both QuantumCumulants and QuantumOptics as backends to simulate system dynamics. 

I want to manipulate vectors and matrices of operators, for example the following code could create a vector containing quadrature operators from a vector of creation and annihilation operators.

```
using QuantumSymbolics 
X = [Destroy; Create]
A = 1/sqrt(2)*[1 1; -1im 1im]  
Y = A*X
```

however this throws the error
```
ERROR: MethodError: no method matching zero(::Type{QuantumInterface.AbstractOperator})
```

I can work around this by defining a new method for zero() like so:
```
import QuantumInterface
import Base.zero
function zero(::Type{QuantumInterface.AbstractOperator})
    return 0.0*I
end
```

I am curious to hear if supporting arrays of operators is within the scope of this package. If so, what other methods might need to be defined?

Krastanov:
Thanks for looking into this! Supporting arrays of operators sounds pretty reasonable, but no-one has needed that yet, so I imagine a lot of other stuff is not implemented either. The Array interface in julia is not that well documented (and to an extent it is implementation-defined), so the easiest way forward probably would just be to implement missing methods as they become obvious.

I would love to see where your project goes and would be happy to help, but some amount of development of the library on your end will be necessary. If you feel up to it, feel free to submit a PR with this addition. I would suggest explicitly constructing `SZeroOperator()` instead of running `0*I`. I should be able to merge and release that pretty rapidly. Please make sure to add a test related to the capabilities (not copying the implementation, but testing some emergent property, like the capability to do array products or that `zero(operator_instance) == 0*operator_instance`).

Krastanov:
fixed by #109

PR #109, opened 2025-04-22
added function Base.zero(x::Symbolic{T})

jeffwack:
Addresses [issue 108](https://github.com/QuantumSavory/QuantumSymbolics.jl/issues/108) by extending Base.zero()

I added tests for the kinds of matrix/vector I was immediately interested in.

I was not able to run the full suite of tests locally due to an error regarding number of threads, but I checked that the isequal() statements evaluated as true in my environment.

>      Testing Running tests...
>      ERROR: julia: -t,--threads=<n>[,auto|<m>]; n must be an integer >= 1
>      ERROR: Package QuantumSymbolics errored during testing

### DIFF
--- a/src/QSymbolicsBase/basic_ops_homogeneous.jl
+++ b/src/QSymbolicsBase/basic_ops_homogeneous.jl
@@ -41,7 +41,8 @@ function Base.:(*)(c::U, x::Symbolic{T}) where {U<:Union{Number, Symbolic{<:Numb
     end
 end
 
-function Base.zero(x::Symbolic{T}) where T
+
+function Base.zero(x::Symbolic{T}) where T<:QObj
     return SZero{T}()
 end
 
--- a/src/QSymbolicsBase/basic_ops_homogeneous.jl
+++ b/src/QSymbolicsBase/basic_ops_homogeneous.jl
@@ -41,11 +41,6 @@ function Base.:(*)(c::U, x::Symbolic{T}) where {U<:Union{Number, Symbolic{<:Numb
     end
 end
 
-
-function Base.zero(x::Symbolic{T}) where T<:QObj
-    return SZero{T}()
-end
-
 Base.:(*)(x::Symbolic{T}, c::Number) where {T<:QObj} = c*x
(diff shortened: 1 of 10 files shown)


src/QSymbolicsBase/basic_ops_homogeneous.jl
```diff
@@ -40,6 +40,11 @@ function Base.:(*)(c::U, x::Symbolic{T}) where {U<:Union{Number, Symbolic{<:Numb
         SScaled{T}(c, x)
     end
 end
+
+function Base.zero(x::Symbolic{T}) where T
```
Krastanov, line 44: This is a type piracy because you are adding a method to a function we do not own (zero) for a type we do not own (Symbolic). You probably want to do this for `QObj`

Krastanov (changes requested):
Can this be achieved by working only on our types, not types from other libraries? See inline comment.

Krastanov:
Feel free to just push here to run the tests on the cloud. We can separately look into what issue is causing them to not run locally.

jeffwack:
Would it no longer be type piracy if I specify that T is a subtype of QObj? 

```
function Base.zero(x::Symbolic{T}) where T<:QObj
    return SZero{T}()
end
```

I tried to define this function in analogy to the function above it which defines multiplication by a scalar

Krastanov:
yup, this looks great

The doctest failures seem to be due to this recent fix to formatting (not sure why we did not catch it there): 

If you have the time to fix the test failures, that would be great. If not, I will get around to it later this week and merge this.

Similarly, if you can move the new piece of code to the bottom of https://github.com/QuantumSavory/QuantumSymbolics.jl/blob/main/src/QSymbolicsBase/literal_objects.jl where SZero is defined, that would be helpful to keep things organized. But if you do not have the time to get to it, later this week I should be able to take care of it.

Thanks again, this is very helpful!

Krastanov:
I forgot to actually link to the source of the doctest failure https://github.com/qojulia/QuantumOpticsBase.jl/pull/195

Krastanov:
This looks great, thank you! The failures are in the downgrade and in the julia alpha version, not really related to your code and non-blocking. I will merge this shortly! It is very much appreciated!

PR #111, opened 2025-04-26
Convert from QuantumCumulants to QuantumOptics

jeffwack:
This adds a file containing a function that converts QuantumCumulants Hamiltonians to QuantumOptics Hamiltonians for bosonic modes and two-level systems. I was a bit lost as to where the proper place for this code is (I suppose the Jaynes-Cummings file should be made into a test?), as well as how to setup the extension module, so I just plopped them down together so you can see the code and let me know how to proceed.

Notes:
-The operators of QuantumCumulants (QC) have names, and these are lost when converting to QuantumOptics (QO).
-QC operators carry a field 'hilbert' which is analogous to what one gets when calling basis() on a QO operator, but the Hilbert spaces of QC are named while the subspaces of a composite space in QO are identified only by their position in the tensor product.
-QC has a nice way to construct operators on a composite space with the idea of 'acts_on,' does QO have something analogous? I am currently tensoring lists with lots of identity operators.
-At the end of the day this conversion works by implementing a lookup table of sorts in the if/else clauses of op_creator(). Is there any way around this? It will result in lots more lines of lookup to extend to all the operators supported in both packages and will require updates when each package expands.

Let me know how you envision the express() function carrying out this conversion. Is the 'target backend' passed to express as an argument? How would you like to deal with the loss of name data passing from QC to QO? I could imagine it being useful to return a dictionary which maps the symbol names of the operators to their QO counterpart.

The other direction of conversion, from QO to QC, raises similar questions. How do we sensibly invent names for the operators? Is it the user's responsibility to provide a dictionary which gives names to all the operators?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 # News
 
-## v0.4.7 - 2025-02-25
+## v0.4.9 - 2025-04-23
+
+- `Base.zero` implemented for quantum symbolic objects.
+
+## v0.4.8 - 2025-02-25
 
 - Proper expression of outer products in QuantumOptics.
 
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "QuantumSymbolics"
 uuid = "efa7fd63-0460-4890-beb7-be1bbdfbaeae"
 authors = ["QuantumSymbolics.jl contributors"]
-version = "0.4.8"
+version = "0.4.9"
 
 [deps]
 Latexify = "23fbe1c1-3f47-55db-b15f-69d7ec21a316"
--- a/docs/src/express.md
+++ b/docs/src/express.md
@@ -6,7 +6,7 @@ DocTestSetup = quote
 end
 ```
 
-A principle feature of `QuantumSymbolics` is to numerically represent symbolic quantum expressions in various formalisms using [`express`](@ref). In particular, one can translate symbolic logic to back-end toolboxes such as [`QuantumOptics.jl`](https://github.com/qojulia/QuantumOptics.jl) or [`QuantumClifford.jl`](https://github.com/QuantumSavory/QuantumClifford.jl) for simulating quantum systems with great flexibiity.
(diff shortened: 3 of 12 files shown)


jeffwack:
Also, just a git question: Is it proper that all the commits from the last PR are seen on this PR? I synced my fork before making the new commit and PR, but GitHub still says I am 8 commits ahead and not just 1 as I was expecting. Not sure what is going on here.

akirakyle:
> Also, just a git question: Is it proper that all the commits from the last PR are seen on this PR? I synced my fork before making the new commit and PR, but GitHub still says I am 8 commits ahead and not just 1 as I was expecting. Not sure what is going on here.

It's because the commits in your previous PR were sqashed then merged. Then your fork was synced via a merge commit of QuantumSymbolics/main thus the two histories are now both there on the main branch of your fork. Github's UI isn't so good at indicating this. Here's what it looks like for me in magit which might help visually clarify.

Screenshot 2025-04-26 at 5 46 09 PM

If you want your fork to reflect the same history as QuantumSymbolics/main than you should be rebasing on top of QuantumSymbolics/main rather than merging. However it shouldn't matter for PRs as I'm pretty sure the squash and merge strategy will only add a commit with your changes here and no any commits in the divergent history.

akirakyle:
> I was a bit lost as to where the proper place for this code is

It doesn't appear this uses anything from QuantumSymbolics? If so then I think either QuantumOptics, QuantumOpticsBase, or QuantumCumulants?

> I suppose the Jaynes-Cummings file should be made into a test?

It looks like it's more a useful example rather than a test in which case it could maybe live in https://github.com/qojulia/QuantumOptics.jl-examples?

> QC has a nice way to construct operators on a composite space with the idea of 'acts_on,' does QO have something analogous?

Not currently, but this is something that has come up before and I think would be good to add and is relevant to https://github.com/qojulia/QuantumInterface.jl/issues/40#issuecomment-2529864881.

> Let me know how you envision the express() function carrying out this conversion.

I just posted a similar issue before I saw this PR and this use case presents another example worth discussing there: https://github.com/qojulia/QuantumInterface.jl/issues/56

jeffwack:
After looking more carefully at the QuantumCumulants documentation, [this functionality is already provided by their package!](https://qojulia.github.io/QuantumCumulants.jl/stable/api/#QuantumCumulants.to_numeric) Oops, though I had some fun re-implementing this.

PR #112, opened 2025-05-11
Fix `Base.show` methods and add corresponding test suite

### LINKED ISSUE #110: Display of SScaledOperator containing SAddOperator
When playing around with making linear combinations of operators I've run into some confusing output.

```
@op A; @op B
C = 2*(A+B)
```
outputs to the REPL

>     2A+B 

After investigating the object C, I can see that the new operator is correct 'under the hood,' it is just improperly displayed. It would be good to see parenthesis on the output, my first guess would be this is every time there is a SScaledOperator where the obj field contains an SAddOperator.

apkille:
Fixes #110 and also the following display for the application of tensor products of operators onto tensor products of bras:
```
julia> (b₁ ⊗ b₂) * (A ⊗ B) # before
(⟨b₁|⟨b₂|)(A⊗B)

julia> (b₁ ⊗ b₂) * (A ⊗ B) # after
⟨b₁|⟨b₂|(A⊗B)
```
Adds a test suite to systematically test/track `Base.show` methods, in addition to the doc tests currently in place.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # News
 
+## v0.4.10 - 2025-05-11
+
+- Polish `Base.show` methods for application products and scaled quantum objects.
+- Add test suite for `Base.show` methods of quantum symbolic types.
+
 ## v0.4.9 - 2025-04-23
 
 - `Base.zero` implemented for quantum symbolic objects.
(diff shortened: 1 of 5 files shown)


Krastanov:
Thanks! Looks good, please update the changelog and merge (and feel free to make a new release).

The downgrade failure is because older version of julia do not print QuantumOptics objects well. The nightly seems unrelated.

PR #113, opened 2025-05-11
Add new symbolic bosonic operators and states

apkille:
I will add `express` functionalities for QuantumOptics and Gabs in separate PRs to make review easier. [This review paper](https://journals.aps.org/rmp/pdf/10.1103/RevModPhys.84.621) is a good quick reference for the quantum objects implemented here.

One thing to note: I would like to add symbolic position and momentum operators in the future. With the way the internals of QSymbolics are currently defined, this would require having support for `PositionBasis` and `MomentumBasis` types in QOBase.jl (https://github.com/qojulia/QuantumOpticsBase.jl/blob/c9c24f96ed9b79a2cb2eeeeb6fc497defedea6a5/src/particle.jl#L6-L63). I suppose we could migrate `ParticleBasis` subtypes into QuantumInterface.jl so that they could be used here? Would that be easy to do? Linking this issue https://github.com/qojulia/QuantumInterface.jl/issues/33.


src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -26,6 +26,21 @@ end
...
 
+"""Two-mode squeezed vacuum state, or EPR state, in defined Fock basis."""
+@withmetadata struct EPRState <: SpecialKet
+    z::Number
+    basis::CompositeBasis
+    function EPRState(z::Number, basis::CompositeBasis)
+        bases = basis.bases
+        length(bases) == 2 && all(x -> x isa FockBasis, bases) || 
+                throw(ArgumentError(lazy"The underlying basis for an EPR state must be a
+                                        tensor product of two bases for single-mode quantum systems."))
```
Krastanov, line 37: This might need to be a single line. Otherwise it prints a bit weird

image

src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -26,6 +26,21 @@ end
 SqueezedState(z::Number) = SqueezedState(z, inf_fock_basis)
 symbollabel(x::SqueezedState) = "0,$(x.z)"
 
+"""Two-mode squeezed vacuum state, or EPR state, in defined Fock basis."""
+@withmetadata struct EPRState <: SpecialKet
```
Krastanov, line 30: `EPR` might be confused with a qubit Bell pair. How about `TwoSqueezeState` or `CVBellPair` or something like that.

src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -26,6 +26,21 @@ end
...
+    basis::CompositeBasis
+    function EPRState(z::Number, basis::CompositeBasis)
+        bases = basis.bases
+        length(bases) == 2 && all(x -> x isa FockBasis, bases) || 
+                throw(ArgumentError(lazy"The underlying basis for an EPR state must be a
+                                        tensor product of two bases for single-mode quantum systems."))
+        return new(z, basis)
+    end
+end
+EPRState(z::Number) = EPRState(z, inf_fock_basis^2)
```
Krastanov, line 41: I am only now noticing, but `SpecialKet` might be a bit of a broken implementation. This and other `SpecialKet` subtypes are marked as `isexpr()=false` while they are definitely expressions, not leaf nodes. No need to fix it in this PR, but this is an issue we should probably track.

src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -161,4 +180,42 @@ julia> qsimplify(S*vac, rewriter=qsimplify_fock)
     basis::FockBasis
 end
 SqueezeOp(z::Number) = SqueezeOp(z, inf_fock_basis)
-symbollabel(x::SqueezeOp) = "S($(x.z))"
\ No newline at end of file
+symbollabel(x::SqueezeOp) = "S($(x.z))"
+
+"""Thermal bosonic state in defined Fock basis."""
+@withmetadata struct ThermalState <: AbstractSingleBosonOp
```
Krastanov, line 186: Similar to the question about EPRPair name, this can also lead to confusion. There are thermal states of qubits and other things. Maybe `CVThermalState` or `BosonicThermalState` or something else that conveys that information. Probably makes sense to sync the name with the one for EPRPair.

src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -161,4 +180,42 @@ julia> qsimplify(S*vac, rewriter=qsimplify_fock)
...
+    photons::Int
```
Krastanov, line 187: This should probably be a real number. But what if someone wants a symbolic one?

src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -161,4 +180,42 @@ julia> qsimplify(S*vac, rewriter=qsimplify_fock)
...
+    basis::FockBasis
```
Krastanov, line 188: another issue that I am only now noticing that should be fixed in the rest of these operators and states is that `basis` should not be a slot in the struct. Only one single basis makes sense here, `inf_fock_basis`. We should be defining the `basis` method, not adding a basis slot. We should track this issue.

src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -161,4 +180,42 @@ julia> qsimplify(S*vac, rewriter=qsimplify_fock)
...
+
+"""Two-mode squeezing operator in defined Fock basis."""
+@withmetadata struct TwoSqueezeOp <: AbstractTwoBosonOp
+    z::Number
+    basis::CompositeBasis
+    function TwoSqueezeOp(z::Number, basis::CompositeBasis)
+        bases = basis.bases
+        length(bases) == 2 && all(x -> x isa FockBasis, bases) || 
+                throw(ArgumentError(lazy"The underlying basis for a two-mode squeeze operator must be a
+                                        tensor product of two bases for single-mode quantum systems."))
```
Krastanov, line 201: same issue with printing due to the new line

src/QSymbolicsBase/predefined_fock.jl
```diff
@@ -161,4 +180,42 @@ julia> qsimplify(S*vac, rewriter=qsimplify_fock)
...
+
+"""Two-mode beamsplitter operator in defined Fock basis."""
+@withmetadata struct BeamSplitterOp <: AbstractTwoBosonOp
+    transmit::Number
+    basis::CompositeBasis
+    function BeamSplitterOp(transmit::Number, basis::CompositeBasis)
+        bases = basis.bases
+        length(bases) == 2 && all(x -> x isa FockBasis, bases) || 
+                throw(ArgumentError(lazy"The underlying basis for a beam splitter operator must be a
+                                        tensor product of two bases for single-mode quantum systems."))
```
Krastanov, line 216: same issue with printing

src/QSymbolicsBase/rules.jl
```diff
@@ -11,7 +11,7 @@ function hasscalings(xs)
     end
 end
 _isa(T) = x->isa(x,T)
-_isequal(obj) = x->(x==obj)
+_isequal(obj) = x->isequal(x, obj)
```
Krastanov, line 14: Why not just `isequal(obj)` -- it provides currying by default?

src/QSymbolicsBase/predefined_fock.jl
Krastanov: as with the previous one, no need to fix it in this PR, but we should track it

Krastanov (changes requested):
Thanks! A few comments below:

- I noticed two separate pre-existing problems with the `predefined_fock` stuff. No need to fix them here, but could you make an issue for each to track them?
- Some questions about how things are printed, how things are named, and a few type constraints.

Krastanov:
migrating ParticleBasis makes sense to me

src/QSymbolicsBase/rules.jl
apkille: I'm not sure if I'm following. Currying isn't supported in Julia, right?

apkille: Wow, this is quite the issue it seems: https://github.com/JuliaLang/julia/pull/24990. I've never seen a Github issue for Julia with that much activity before 😄

Krastanov: Indeed, but some functions have manually defined curried methods. `isequal(a)(b)` computes `isequal(a, b)`. and `==(a)(b)` computes `a==b`.

Krastanov: Yeah, that issue is frequently called "the issue that shall not be named" or "the issue that the dev team does not want to touch even with a 10ft pole". At this point it should probably just get closed. There are currying macros that are pretty ok -- it does not need to be in the base language (but maybe a base library would be nice).

apkille:
@krastanov let me know when this looks good to merge.

apkille:
@Krastanov ping.

Krastanov (approved):
Sorry for the lateness on reviewing this. If the changes I made in the last commit make sense and are fine with you, and if the tests pass, feel free to merge.

Issue #116, opened 2025-05-18
Conversion from symbolic object to numeric objects from the QuantumToolbox.jl library (by implementing a backend for the `express` API)


Krastanov:
One of the main value-adds of QuantumSymbolics.jl is to enable converting symbolic expressions into the appropriate numerical representation for use with a variety of simulators. We have such capabilities withe respect to a few simulators now, e.g. converting the symbolic `Z1` into either a Ket (from `QuantumOptics.jl`) or a Stabilizer tableau (from `QuantumClifford.jl`). Here is an example:

```
julia> using QuantumSavory, QuantumOpticsBase, QuantumClifford

julia> express(X1, QuantumOpticsRepr())
Ket(dim=2)
  basis: Spin(1/2)
 0.7071067811865475 + 0.0im
 0.7071067811865475 + 0.0im

julia> express(X1, CliffordRepr())
𝒟ℯ𝓈𝓉𝒶𝒷
+ Z
𝒮𝓉𝒶𝒷
+ X
```

### This bounty is on introducing symbolic-to-numeric conversion capabilities for the `QuantumToolbox.jl` library

About `QuantumToolbox.jl`: `QuantumToolbox.jl` is a recent alternative to `QuantumOptics.jl` with very similar capabilities. It would be useful for crosschecks and overall ecosystem health to support that library as well. A big difference between the two libraries is that `QuantumOptics.jl` enforces a notion of a semantic basis (e.g., a Fock basis is different from a spin basis of the same number of dimmensions), while `QuantumToolbox.jl` has a much flatter inner structure for its states and operators. Nonetheless, given the significant similarities, one can structure the implementation very similarly to the one for `QuantumOptics`.

The most useful references would be:

- [what is the `express` API that is used for conversion from symbolics to numerics](https://doc.quantumsavory.org/QuantumSymbolics.jl/dev/express/)
- [how `express` is implemented for `QuantumOptics`](https://github.com/QuantumSavory/QuantumSymbolics.jl/tree/main/ext/QuantumOpticsExt)
- [the overall documentation of QuantumSymbolics](https://doc.quantumsavory.org/QuantumSymbolics.jl/dev/)
- [conditionally loaded package extensions in julia](https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)) which is how we have structured the dependencies on various external libraries in QuantumSymbolics

For this bounty to be completed, one would need:

[...]

PR #121, opened 2025-05-30
start of QuantumToolboxExt

jeffwack:
This is incomplete, but I have made some progress on #116 and have some questions. My strategy was to copy the file QuantumOpticsExt and go line by line to convert to QuantumToolbox. I ran into name collisions (basis and \otimes) come to mind, so I decided to be conservative and use the QuantumToolbox.function() form for all function calls.

I ran into a roadblock converting this:
```
const xyzopdict = Dict(:X=>_x, :Y=>_y, :Z=>_z)
const xyzstatedict = Dict(:X=>(_s₊,_s₋),:Y=>(_i₊,_i₋),:Z=>(_l0,_l1))
for control in (:X, :Y, :Z)
    for target in (:X, :Y, :Z)
        k1, k2 = xyzstatedict[control]
        o = xyzopdict[target]
        gate = QuantumToolbox.tensor(proj(k1),_id) + QuantumToolbox.tensor(proj(k2),o)
        structname = Symbol(control,"C",target,"Gate")
        let gate=copy(gate)
            @eval express_nolookup(::$(structname), ::QuantumToolboxRepr) = $gate
        end
    end
end
```

Specifically, you cannot copy() a QuantumToolbox operator. This example shows the error:
```
using QuantumToolbox
copy(sigmap())
```

So, to avoid type piracy we cannot implement the conversion as written without contributing a method for copy() to QuantumToolbox?

Another route would be to rework that chunk to avoid the copy, but I am not so clear on how it does what it does. Why the macro?


apkille:
> Specifically, you cannot copy() a QuantumToolbox operator. This example shows the error:

Could you just print the error here? It makes review and answering questions faster.

> Another route would be to rework that chunk to avoid the copy, but I am not so clear on how it does what it does. Why the macro?

Notice that the for loops are iterating through `Symbol`s. [Here's a discussion on](https://discourse.julialang.org/t/eval-vs-eval/17091/7.) what `@eval` does that probably is helpful.

> So, to avoid type piracy we cannot implement the conversion as written without contributing a method for copy() to QuantumToolbox?

Yeah, if you do want a new `copy` method for the QObjs, I would recommend submitting a PR to QuantumToolbox rather than committing type piracy here. Although that's really not necessary, there's a bunch of different ways you could define the `express_nolookup` methods here. To help understand what the chunk does, note how we do it in actually defining symbolic objects like (`XCXGate` and `ZCYGate`) in the source code:
https://github.com/QuantumSavory/QuantumSymbolics.jl/blob/a4980ef8c691ad09c621c79b1cd0fe2532d6289c/src/QSymbolicsBase/predefined.jl#L119-L132
When defining the conversion in QuantumOptics, we are looping through each permutation of the two qubit gates and lining them up with the spin operators defined in QuantumOptics.

ytdHuang:
@jeffwack @apkille

Just an update to you, we have released `QuantumToolbox v0.32` (see our [ChangeLog](https://qutip.org/QuantumToolbox.jl/stable/resources/changelog))

apkille:
@jeffwack what's the status on this PR? I can help you finish it if you want.

jeffwack:
@apkille That sounds great to me! Apologies, I let this fall off my todo list. If you want to take this over that is fine by me :)

Krastanov:
this seems related https://github.com/qutip/QuantumToolbox.jl/pull/470

apkille:
Shall we merge this and create issues for the failing tests and deal with them in future PRs? I can do that this week during JuliaCon.

Krastanov:
I think this will be causing actual issues to some folks. I would not be surprised if this issue pops up in a lot of environments that use QuantumSavory.jl. I would prefer to wait at least for a comment from the QuantumToolbox folks, because this might be unfixable without their commitment, and I would prefer not to have a future PR removing this extension.

apkille:
Ok, sounds good

ytdHuang:
Ahh, the JuliaRegistry Bot is still running lol

Krastanov:
The macos test is just a flaky test

PR #129, opened 2025-07-03
Add Gaussian channel support for Gabs

apkille:
This is a small PR for translating attenuation and amplification Gaussian channels supported in Gabs. [This issue](https://github.com/apkille/Gabs.jl/issues/59) is tracking the implementation of more channels in Gabs; when they're supported, I will ensure that symbolic representations will be added here.

### DIFF
--- a/src/QSymbolicsBase/predefined_CPTP.jl
+++ b/src/QSymbolicsBase/predefined_CPTP.jl
@@ -1,8 +1,34 @@
-export PauliNoiseCPTP, DephasingCPTP, DephasingCPTP, GateCPTP
+export PauliNoiseCPTP, DephasingCPTP, DephasingCPTP, GateCPTP, AttenuatorCPTP, AmplifierCPTP
 
 abstract type NoiseCPTP <: Symbolic{AbstractSuperOperator} end
 isexpr(::NoiseCPTP) = false
 basis(x::NoiseCPTP) = x.basis
+
+"""
+    AttenuatorCPTP(theta::Number, noise::Int)
+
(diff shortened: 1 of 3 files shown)


src/QSymbolicsBase/predefined_CPTP.jl
```diff
@@ -1,9 +1,35 @@
...
 
+"""
+    AttenuatorCPTP(theta::Number, noise::Int)
+
+Attenuation CPTP map, defined by the beam splitter rotation parameter `theta`
+and thermal noise parameter `noise`.
+"""
+@withmetadata struct AttenuatorCPTP <: NoiseCPTP
+    theta::Real
+    noise::Int
```
Krastanov, line 15: I am surprised `noise` is Int. Is this on purpose?

Krastanov (commented):
I have a question about the Int type, but otherwise this seems good to merge.

src/QSymbolicsBase/predefined_CPTP.jl
apkille: I suppose we could instead parametrize `AttenuatorCPTP` in terms of `photons` or `nphotons` to indicate that  the mode that we're operating on is coupled to a bosonic thermal state with `(n)photons`. Usually though when I see mention of this channel though it's in terms of `noise = 2*nphotons + 1`.

Krastanov: But I would have thought `nphotons` is average number of photons, so it does not need to be an int?

Issue #67, opened 2024-07-09
All struct fields should be (concretely) typed (potentially with a type parameter) [$200]


Krastanov:
<details>
<summary><strong>Bug bounty logistic details</strong> (click to expand)</summary>

To claim exclusive time to work on this bounty either post a comment here or message [skrastanov@umass.edu](mailto:skrastanov@umass.edu) with:

- your name
- github username
- **(optional)** a brief list of previous pertinent projects you have engaged in

Currently the project is claimed by `no one` until `...`.

If you want to, you can work on this project without making a claim, however claims are encouraged to give you and other contributors peace of mind. Whoever has made a claim takes precedence when solutions are considered.

You can always propose your own funded project, if you would like to contribute something of value that is not yet covered by an official bounty.
</details>

# Project: All struct fields should be (concretely) typed (potentially with a type parameter) [$200]

Many of the symbolic types in this library are rather clumsily defined, without much attention being paid to the types of their fields. The goal of this bounty is to specify them and parameterize them as much as possible and reasonable, in preparation for next steps like the use of sum types and Unityper.jl. [This forum discussion might provide useful background](https://discourse.julialang.org/t/when-is-it-better-to-use-non-concrete-types-in-structs/29041/11).

**Required skills**: Understanding of Symbolics.jl and Julia's type system

**Reviewer**: Stefan Krastanov

**Duration**: 1 month

#### Payout procedure:

The Funding for these bounties comes from the National Science Foundation and from the NSF Center for Quantum Networks. The payouts are managed by the NumFOCUS foundation and processed in bulk once every two months. If you live in a country in which NumFOCUS can make payments, you can participate in this bounty program.

[Click here for more details about the bug bounty program.](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

R-ohit-B-isht:
@Krastanov can i take this?

Krastanov:
Hi, Rohit! Do you have past experience with the julia language? This is a somewhat idiomatic topic that would be easy for someone knowing the language, but not obvious otherwise.

thofma:
Is someone working on this? If not, I would like to take a stab at this.

Krastanov:
Pardon the slow response, last week was a summer workshop. This one is indeed still open and unclaimed. A couple of notes though:

- I have a review backlog from last week so I might be a bit slow to provide reviews
[...]

PR #190, opened 2026-05-12
Concretely type fock and scalar expression fields

Trivolous:
## Summary
- parameterize scalar expression wrapper fields so their stored coefficient/object/vector types are concrete per instance
- parameterize Fock-space states and Gaussian/Fock operators that previously stored `Number` fields
- keep public constructors such as `CoherentState(im)` and `PhaseShiftOp(pi)` working by forwarding to the inferred parametric type

## Bounty scope
This is a focused, reviewable slice of #67 rather than an attempt to finish the full ADT/Moshi direction in one PR.

## Verification
- `git diff --check`

I could not run the Julia test suite locally because `julia` is not installed in this Codex workspace.

Refs #67

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # News
 
+## unreleased
+
+- Concretely type scalar expression wrappers and Fock-space numeric symbolic fields.
+
 ## v0.4.16 - 2026-04-01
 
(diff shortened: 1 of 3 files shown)


Krastanov:
banned and reported for autogenerating broken PRs

PR #217, opened 2026-07-31
All struct fields are concretely typed

### LINKED ISSUE #67: All struct fields should be (concretely) typed (potentially with a type parameter) [$200]

AJ0070:
Closes #67.

Every field of every struct now has a declared type that is concrete for any concrete
instantiation, either directly (`idx::Int`, `terms::Vector{Symbolic{T}}`) or through a type
parameter (`basis::B`, `alpha::T<:Number`, `ket::K<:Symbolic{AbstractKet}`); 52 of the 79 structs
are now parametric and no untyped or `Any` field is left. Fixed-arity children get a type parameter,
while variadic ones keep a concrete container with an abstract element type, so that the type of an
expression never depends on its size. None of this was possible before, because `@withmetadata`
hard-coded a single type parameter (`new{S}(...)`); it now carries over any number of them and adds
`Foo(fields...)` whenever every parameter is spelled out as the type of a field. It stays an *inner*
constructor on purpose: that is what keeps Julia from generating the default ones, whose last
argument would be the metadata. `QuantumSavory` defines `BarrettKokBellPairW(ηᴬ,ηᴮ)` on a one-field
decorated struct, and with an outer constructor that silently overwrites a generated method and
fails precompilation. Equality is deliberately left alone: `isequal`/`hash` compared `typeof`, which
would have made `CoherentState(1) != CoherentState(1.0)`, so they now compare the struct plus the
kind of quantum object (`samekind`/`qobjtype`), keeping the field values as what decides equality
while `zero(ket)` and `zero(bra)` stay distinct.

The stricter types surfaced three latent bugs, each fixed here: `flattenop` spliced *any* `*`-headed
term into an operator product, so a ket and a bra could end up inside `SMulOperator.terms`;
`inv(A)*B` threw a `MethodError`; and `isequal` compared arguments with `zip`, making `A*B` equal to
`A*B*C`. A new `test/general/struct_types_tests.jl` locks the invariant in: no field of any struct
in the package or its extensions may be underspecified, every field of a representative instance
must be concrete, and the corpus must cover *every* struct, so a newly added struct without a
representative fails the suite. The `general` suite goes from 450 pass, 2 broken to 852 pass, 2
broken with no failures, downstream `QuantumSavory` precompiles clean and passes (14153 pass, 8
broken), and `jet` drops from 2 reports to 1, the remaining one being a pre-existing missing `show`
for a sum of superoperators that JET can only now see, left alone since fixing it would turn the
`@test_broken` in `jet_tests.jl` into an unexpected pass. Diffing ~150 objects against `main`
through `show`, `latexify`, `isequal`/`hash`, `qsimplify`, `qexpand`, `basis` and `express` in both
the QuantumOptics and the Clifford representations gives identical output, except that `latexify` of
a sum can order its terms differently, following the changed hash of the literal objects, an order
that comes out of a `Dict` and has never been meaningful, which is why `show` sorts.

- [x] The code is properly formatted and commented.
- [x] Substantial new functionality is documented within the docs. <small>No new user facing functionality; the docstrings of the touched types, of `@withmetadata` and of the helpers in `utils.jl` were updated.</small>
- [x] All new functionality is tested.
- [x] All of the automated tests on github pass. <small>Everything passes locally (`general`, `jet`, the doctests, Aqua, the benchmarks and the downstream `QuantumSavory` suite); to be ticked once CI has run.</small>
- [x] We recently started enforcing formatting checks. If formatting issues are reported in the new code you have written, please correct them.

If you are submitting for a bug bounty:
- [x] I have read and followed the [AI usage and disclosure policy](https://github.com/QuantumSavory/.github/blob/main/BUG_BOUNTIES.md)

AI Disclosure: I used Claude Fabel to initially understand the problem, code scaffolding, adding comments, documentation, and getting full test coverage.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,13 @@
 # News
 
+## v0.4.18 - 2026-08-17
+
+- All the fields of all the symbolic structs are now concretely typed, in most cases through newly added type parameters (e.g. `SKet{B<:Basis}`, `SScaled{T<:QObj,C<:SymCoeff,O<:Symbolic{T}}`, `SProjector{K<:Symbolic{AbstractKet}}`, `CoherentState{T<:Number}`), in preparation for a future move to sum types. The constructors keep their previous signatures and infer the new parameters, but code that spells out the full type of a symbolic object has to be updated.
+- `isequal` and `hash` of symbolic quantum objects ignore the type parameters that merely record the types of the fields, so e.g. `CoherentState(1)` and `CoherentState(1.0)` are still equal.
+- Fix `isequal` considering two products (or tensor products) of different lengths equal, e.g. `A*B` and `A*B*C`.
+- Fix a `MethodError` when multiplying an operator by the inverse of a different operator.
+- A product of operators no longer splices the arguments of other `*`-headed expressions (an outer product or a superoperator application) into the `SMulOperator` it builds, so its terms are always operators. The printed and expressed results are unchanged.
+
 ## v0.4.17 - 2026-08-06
 
 - **(fix)** Complete Clifford observable lowering for symbolic Pauli tensors, identities, and native `PauliOperator`s.
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "QuantumSymbolics"
 uuid = "efa7fd63-0460-4890-beb7-be1bbdfbaeae"
 authors = ["QuantumSymbolics.jl contributors"]
-version = "0.4.17"
+version = "0.4.18"
 
 [workspace]
 projects = ["benchmark", "docs"]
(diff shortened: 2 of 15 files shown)


AJ0070:
@Krastanov gentle ping!

AJ0070:
@Krastanov

PR #224, opened 2026-09-05
Use DocumenterCodeBlocks in documentation

Krastanov-agent:
Summary:
- add DocumenterCodeBlocks 1.x to the documentation environment
- enable CodeBlocks alongside the existing citation plugin
- fail the documentation build before deployment when the plugin reports a referenced docstring-quality issue
- record the rendered documentation and CI behavior in the changelog

Warning behavior:
- DocumenterCodeBlocks emits these diagnostics as ordinary Julia `@warn` records, not Documenter errors
- `warnonly = false` therefore does not escalate them; the explicit post-`makedocs` gate checks the plugin warning state

Validation:
- `JULIA_PKG_SERVER_REGISTRY_PREFERENCE=conservative julia --project=docs -e "using Pkg; Pkg.instantiate()"`
- `julia -tauto --project=docs docs/make.jl` (clean build passes with no `CodeBlocks:` warnings)
- negative check: temporarily removed the `X` signature; the build emitted the expected `CodeBlocks:` warning and exited 1; restored before commit


Krastanov-agent:
CI audit: the two current-package CI legs fail at the same Aqua check for the existing tensor(xs::SymQObj{T}...) where T method at src/QSymbolicsBase/basic_ops_homogeneous.jl:211, which Aqua now reports as having an unbound type parameter. These jobs resolve QuantumInterface 0.4.4; that release migrated tensor to TensorCore and removed the previous zero-argument tensor method. The latest base CI run on 2026-08-27 resolved QuantumInterface 0.4.3 and passed. This PR changes documentation configuration plus the changelog, and DocumenterCodeBlocks depends only on CRC32c, Documenter, and JuliaSyntax, so there is no dependency path from this docs addition to QuantumInterface. A plain rerun with the current resolver state is unlikely to change this result.

The Downgrade failure is also unrelated to the PR diff: the current base SHA already failed identically in its 2026-08-27 Downgrade run (https://github.com/QuantumSavory/QuantumSymbolics.jl/actions/runs/33128049531). Both runs resolve RuntimeGeneratedFunctions 0.5.25 and, during Julia 1.12 coverage precompilation, its PrecompileTools workload reaches generated_callfunc before the generated method is visible in the current world age. The resulting method-too-new error cascades into the reported dependent-package precompile failures. CHANGELOG.md causes this paths-ignore workflow to run, but does not cause its failure; an unchanged rerun is unlikely to help.

